PR #6, opened 2019-08-12
Correct CMAKE_CXX_FLAGS for ninja, use unicode windows APIs explicitly

p-podsiadly:
This solves two problems when using Visual Studio's CMake and Ninja:

* `ninja` complains about new line in `CMAKE_CXX_FLAGS`

* `UNICODE` is not defined, which makes (e. g.) `CreateWindow` an alias for `CreateWindowA` instead of `CreateWindowW` - fixed by using `*W` variants where necessary


djowel:
Looks good!

PR #13, opened 2020-01-11
modern CMake

Xeverous:
- changed global commands to target_* commands with PRIVATE or PUBLIC
properties
- removed unused elements_root variable
- replaced CMAKE_SYSTEM_NAME conditions with names guaranteed by CMake
note that UNIX is also present when APPLE is, so when writing conditions
test for APPLE first or write if (UNIX AND NOT APPLE)
- convention unification: variables uppercase, functions lowercase
- removed globbing (discouraged) and replaced with target_sources
- non-boost dependencies are now discovered using pkg-config with
IMPORTED_TARGET which supports autopropagation of compiler/linker flags
- preserved existing MACOS-specific and MSVC-specific behavior
- some Windows-specific behavior was corrected to MSVC-specific


Xeverous:
Some extra notes:
- the (now removed) glob for *.c had no effect
- `stb_image.h` was not formally added (the glob was only for `*.hpp`) - it was a coincidence some other files had the same include directory which caused it to silently work
- the cmake file which finds Cairo was removed because:
  - it was using pkg-config but the legacy way (which set some `CAIRO_` variables instead of `PkgConfig::cairo` target)
  - some of its functionality is now built in CMake
  - the modern way with pkg-config requries only 1 line to find and 1 line to use as long as the library provides correct `.pc` file

GCC and Clang will output a ton of warnings - expect their fixes in future PRs. Some warnings could actually detect something (seem some "missing initializer for member ..." with braced-init)

And of course let me know if I broke anything - I tested only the Windows build.

Xeverous:
Freetype will also need to be public. It is present in elements headers

https://github.com/cycfi/elements/blob/a5724e68ff446a177523f64701d708d68e1667fe/lib/include/elements/support/detail/canvas_impl.hpp#L9-L19

djowel:
I'll be targeting a different branch for this, so we can have it stabilize smoothly.

djowel:
odd, sometimes github does not show the branches dropdown for changing the target branch :-/ it was there until i added a new branch and refreshed.

Xeverous:
It is shown when you click the button to edit PR title

djowel:
> It is shown when you click the button to edit PR title

Ah right! Silly interface :-P

PR #14, opened 2020-01-11
fix unused variable warnings

Xeverous:
also removed some extra semicolons

### DIFF
--- a/lib/host/windows/base_view.cpp
+++ b/lib/host/windows/base_view.cpp
@@ -184,7 +184,7 @@ namespace cycfi { namespace elements
 
       mouse_button get_button(
          HWND hwnd, view_info* info, UINT message
-       , WPARAM wparam, LPARAM lparam)
+       , WPARAM /* wparam */, LPARAM lparam)
       {
          float pos_x = GET_X_LPARAM(lparam);
          float pos_y = GET_Y_LPARAM(lparam);
@@ -228,8 +228,6 @@ namespace cycfi { namespace elements
                break;
          }
 
-         int click_count = 1;
-
          mouse_button::what which;
          switch (message)
          {
@@ -252,7 +250,7 @@ namespace cycfi { namespace elements
          return {
             down,
             info->click_count,
-            mouse_button::left,
+            which,
             get_mods(),
             { pos_x, pos_y }
          };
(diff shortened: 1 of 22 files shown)


lib/host/windows/base_view.cpp
```diff
@@ -252,7 +250,7 @@ namespace cycfi { namespace elements
          return {
             down,
             info->click_count,
-            mouse_button::left,
+            which,
```
djowel: i haven't checked, but is this the same?

Xeverous:
2 things are suspicious here

https://github.com/cycfi/elements/blob/a5724e68ff446a177523f64701d708d68e1667fe/lib/host/windows/base_view.cpp#L185-L259

- the `click_count` is never used (removed)
- `mouse_button::what which` is set in the switch but never used (now in the function return statement)

djowel:
I find it awkward to weave in and out of github code lists. Can we move code that can potentially change behavior in a separate commit/pr, then we can focus more later? (e.g. click_count and mouse_button::left). There's a lot going on here already. I'd prefer if we keep the changes simple to begin with (i.e. fixes for warnings by commenting out arguments).

djowel:
> I find it awkward to weave in and out of github code lists. Can we move code that can potentially change behavior in a separate commit/pr, then we can focus more on those later? (e.g. click_count and mouse_button::left). There's a lot going on here already. I'd prefer if we keep the changes simple to begin with (i.e. fixes for warnings by commenting out arguments).

So basically, I'd prefer to review changes to code that needs further investigation in a separate PR. So far I see only the click_count and mouse_button::left stuff that I want to review later in another PR. ALl the rest is fine.

lib/host/windows/base_view.cpp
Xeverous: No, this behaviour is different. Previously it was always returning `mouse_button::left`, now it returns `which` which depends on the switch. Before this change, the variable `which` was unused (so the switch was dead code).

Xeverous:
> I'd prefer to review changes to code that needs further investigation in a separate PR

Ok, I will push a reverting commit for the `click_count` and `which` so you can squash-merge this PR with no diff to this code. Without these changes, this PR will have no impact on behavior.

djowel:
Thanks!

PR #20, opened 2020-01-12
CMake improvements

Xeverous:
- if no CMAKE_BUILD_TYPE is specified, it will now default to Release
- if if CMAKE_BUILD_TYPE is Release and LTO is supported, it will be
added to elements library and examples targets
- removed trailing whitesapce


Xeverous:
Enough for today, going to sleep now. Expect more PRs tomorrow.

PR #21, opened 2020-01-12
fixes for type and comparison warnings

Xeverous:


Xeverous:
Few things needs to be checked:

- the change in `canvas.cpp`: `FT_Get_Sfnt_Name(face, i, &name);` was duplicated, by the function doc it should have no effect to call it twice
- I got some UTF-8 warnings in the editor, 1 file was re-saved. Please check this. Also note that `u8` literals in C++20 are `char8_t`, not `char` (breaking change) so consider rewriting them to something else
- loops and comparisons should have no behavioral changes - I would like to change `i`s to `std::size_t` but many of them are used with value `-1` so I added casts instead

Xeverous:
I rebased the commit onto the current `develop` HEAD and removed unwanted string literals changes.

PR #63, opened 2020-02-11
Fix undefined behaviour

omartijn:
The on_focus method was declared to return a gboolean, but did not
actually return anything. This prevented everything from working.

The function is now declared void.


Xeverous:
I haven't checked this on any Linux system. I guess for all issues like this there should be warning in both GCC and Clang. Do you get any more warnings?

omartijn:
This was the only place where a non-void function was not returning anything (resulting in undefined behaviour). I do get a repeated error about the assignment operator for `basic_context` and `context` being implicitly deleted because it has reference members. But this is only a warning as long as you don't actually try to use it.

If I turn -Wall and -Werror on I get screens full of warnings, but they seem to be all about unused parameters and missing field initializers.

Xeverous:
> I do get a repeated error about the assignment operator for basic_context and context being implicitly deleted because it has reference members. But this is only a warning as long as you don't actually try to use it.

Cna you paste how it looks like? From my talks with @djowel the `context` and `basic_context` are such by design and should never be assigned. Having reference members should not trigger any warnings.

> I do get a repeated error about the assignment operator for basic_context and context being implicitly deleted because it has reference members. But this is only a warning as long as you don't actually try to use it.

Majority of them should be fixed in the develop branch. I'm currently working on (hopefully and probably last 2) PRs that solve uninitialized/unused variable warnings.

djowel:
It's good someone is looking at the Linux port. I'll need to spend more time on it, perhaps in the coming days. Thanks @omartijn

djowel:
oops, i missed that his targets 'master' branch. i'll need to merge from master then.

omartijn:
The error I get is the following:

```
/home/martijn/git/cycfi/elements/lib/include/elements/support/context.hpp:32:22: warning: explicitly defaulted copy assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
      basic_context& operator=(basic_context const&) = default;
                     ^
/home/martijn/git/cycfi/elements/lib/include/elements/support/context.hpp:34:30: note: copy assignment operator of 'basic_context' is implicitly deleted because field 'view' is of reference type 'elements::view &'
      elements::view&        view;
                             ^
/home/martijn/git/cycfi/elements/lib/include/elements/support/context.hpp:59:16: warning: explicitly defaulted copy assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
      context& operator=(context const&) = default;
               ^
/home/martijn/git/cycfi/elements/lib/include/elements/support/context.hpp:39:20: note: copy assignment operator of 'context' is implicitly deleted because base class 'cycfi::elements::basic_context' has a deleted copy assignment operator
   class context : public basic_context
                   ^
/home/martijn/git/cycfi/elements/lib/include/elements/support/context.hpp:32:22: note: explicitly defaulted function was implicitly deleted here
      basic_context& operator=(basic_context const&) = default;
                     ^
/home/martijn/git/cycfi/elements/lib/include/elements/support/context.hpp:34:30: note: copy assignment operator of 'basic_context' is implicitly deleted because field 'view' is of reference type 'elements::view &'
      elements::view&        view;
                             ^
```

It happens because we explicitly marked the assignment operator to be defaulted, which is not possible, so the compiler actually deleted it for us. Since we don't actually _use_ the assignment operator it is only a warning. Explicitly `delete`-ing it instead of `default`-ing it would solve these warnings.

BTW: Would you prefer further pull-requests to go to the develop branch?

Xeverous:
Fair warning. These ctors/operators should either not be written or explicitly `delete`d. For PRs use `develop` branch, so far @djowel wanted them all on this branch and almost all of my PRs were merged into develop.

djowel:
It is deleted in develop branch as it should be:
https://github.com/cycfi/elements/blob/develop/lib/include/elements/support/context.hpp#L59

We'll soon merge to master as things stabilize.

djowel:
> These ctors/operators should either not be written or explicitly deleted.

Oh..., it's OK for the copy ctor, but not for assign. It's perfectly fine to copy construct a context.

PR #64, opened 2020-02-12
Fix linking of freetype and threads on linux

omartijn:
There were two issues:

- Checking for LINUX instead of UNIX platform
- pkg-config for freetype did not work

For the second I have changed the code to use find_package instead. To
be able to use proper CMake targets for this I had to bump the required
version number. Please let me know if this is not acceptable.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -197,9 +197,9 @@ if (MSVC)
    else()
 	  target_link_libraries(elements PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/external/freetype/win32/freetype.lib)
    endif()
-elseif (LINUX OR WIN32)
-   pkg_check_modules(freetype REQUIRED IMPORTED_TARGET freetype)
-   target_link_libraries(elements PUBLIC PkgConfig::freetype)
+elseif (UNIX OR WIN32)
+   pkg_check_modules(freetype2 REQUIRED IMPORTED_TARGET freetype2)
+   target_link_libraries(elements PUBLIC PkgConfig::freetype2)
 endif()
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


Xeverous:
I agree with `LINUX` to `UNIX` change - already did this in previous PRs but apparently missed few of them.

For the `find_package`, this is a breaking change for me on Windows. I prefer to use pkg-config there because it is more stable and less reliant on complex CMake settings. What was the error you got?

At worst, we could just add an `option` that lets to choose at CMake invokation time how freetype should be searched.

omartijn:
The error I got was this:

```
-- Checking for module 'freetype'
--   No package 'freetype' found
```

It works for me if I have pkg-config search for `freetype2` instead of `freetype`, strangely enough, the package is then still called `PkgConfig::freetype`. It seemed easier to simply use `FindFreetype` instead. Note that this is _only_ changed on UNIX-like systems.

The only thing is that `FindFreetype` only exports targets from CMake 3.10, which means a version bump.

Xeverous:
> It works for me if I have pkg-config search for freetype2 instead of freetype, strangely enough, the target is then still called PkgConfig::freetype

Things you probably were not aware of:

- When you do `pkg_check_modules(foo REQUIRED IMPORTED_TARGET bar)` it will always appear as `PkgConfig::bar` target. The last string is only the name of the target and the first string is the actual `.pc` file that is being searched.
- Freetype has multiple versioning systems running simultaneously. You can find them in their's repo changelog. For backwards compatibility, `freetype` is still available after installation but it points into `freetype2` library.
- For the reason above, I have actually both `freetype.pc` and `freetype2.pc` files after installing the library. They both point to the same library object.

> It seemed easier to simply use FindFreetype instead. Note that this is only changed on UNIX-like systems.

It is not. You have edited it under `elseif (UNIX OR WIN32)`

___

So ... I think you can just change the `freetype` to `freetype2` in `pkg_check_modules`. This will remove the requirement to bump minimum CMake version.

djowel:
@Xeverous feel free to merge this once you are comfortable with it.

omartijn:
I've reverted the CMake version bump and use of FindFreetype.

Xeverous:
I confim it builds successfully. @djowel I have no merge priviledge, you will need to do it yourself.

djowel:
Done.

djowel:
Many thanks @omartijn and @Xeverous

Issue #75, opened 2020-03-23
build with MinGW and GTK ?


yslking:
I noticed it is really easy to build elements with MinGW and GTK. I simply made everything behave as Linux except for the boost::asio and it works fine. Are there any plans to officially support building for MinGW and GTK?

djowel:
> I noticed it is really easy to build elements with MinGW and GTK. I simply made everything behave as Linux except for the boost::asio and it works fine. Are there any plans to officially support building for MinGW and GTK?

That's good! Sure we can do that. Are you willing to do a PR on the Setup and Installation instructions here: http://cycfi.github.io/elements/setup ?

Xeverous:
What are the potetial benefits/drawbacks of such build configuraton? Does GTK on Windows bring some extra functionality over current plain WinAPI implementation?

yslking:
> What are the potential benefits/drawbacks of such a build configuration? Does GTK on Windows brings some extra functionality over current plain WinAPI implementation?
@Xeverous 
It will not replace plain WinAPI implementation. I hope to add an option rather than replace anything. And adding support for GTK on Windows will not bring extra complications for the project.
 
benefits:
(Compared to MSVC) Many people (I guess) and I are not familiar with msvc's error message.
(Compared to MFC GCC as you said) GCC with WinAPI is not well implemented, as you posted in issue #9, but GTK for Windows supports new features like GetWindowDpi well. And GTK might support windows 7 for you.

(pseudo-)benefit:
It is just a normal choice to support it, as many other opensource cross-platform softwares(e.g. GIMP, inkscape and Anjuta) do.

drawbacks:
It may be slower in runtime and compile time.
 The multiple reference issue of GCC (I remembered it was put forward by you, but I am not able to find it now. I am sorry.)

yslking:
> That's good! Sure we can do that. Are you willing to do a PR on the Setup and Installation instructions here: http://cycfi.github.io/elements/setup ?

Sure. I am glad to do it several days later.

Xeverous:
> The multiple reference issue of GCC (I remembered it was put forward by you, but I am not able to find it now. I am sorry.)

Did you mean #69? This is unrelated to GTK, it's a MinGW LTO bug.

yslking:
thanks, it is #69. I knew it is unrelated to GTK. I mentioned it  because I thought P(use MinGW | use GTK) > P(use MinGW | use win32 api).

[...]

PR #76, opened 2020-03-23
support building with mingw and gtk

yslking:
see issue #75 

What did I change:
1. Create a CMake cached variable `WINDOWS_GRAPHICS_LIBRARY`. "native" means build it with Win32 APIs, and "gtk" means build it with Gtk. Default to "native".
2. Create a preprocessor macro `__WINDOWS_WITH_GTK__` if `WINDOWS_GRAPHICS_LIBRARY` is set to "gtk".
3. Modify several lines of source codes to support Windows + Gtk.
4. msvc do not will support utf-8-no-bom files. following lines will deal with it.
5. Update setup document.
```
add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>") 
add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>") 
```

What did I test:
1. MinGW + Gtk
2. msvc(NMake Makefiles) 
3. Linux

What didn't I test:
1. macOS. (I do not have a MacBook. )
2. MinGW + Win32 APIs. (Same error as before the change. i.e. cannot find some functions about folders and HiDPI).


CMakeMain.txt
```diff
@@ -70,7 +70,7 @@ if (APPLE)
       ${ELEMENTS_RESOURCES}
       ${ELEMENTS_APP_RESOURCES}
    )
-elseif (UNIX)
+elseif (UNIX OR (WINDOWS_GRAPHICS_LIBRARY STREQUAL "gtk"))
```
djowel: I think this should be named WINDOWS_UI_LIBRARY, but better yet, can we have a generic one for all targets, like perhaps HOST_UI_LIBRARY, which can be "cocoa" (for MacOS), "gtk" (for Linux and Windows), and "win32" (for Windows)?

CMakeLists.txt
```diff
@@ -27,6 +27,15 @@ if (NOT DEFINED CYCFI_JSON_ROOT)
   set(CYCFI_JSON_ROOT ../json)
 endif()
 
+set(WINDOWS_GRAPHICS_LIBRARY "native" CACHE STRING "build with windows native gui api or gtk")
+if(WINDOWS_GRAPHICS_LIBRARY STREQUAL "gtk")
+    add_definitions(-D__WINDOWS_WITH_GTK__)
```
djowel: Avoid double underscores, they are reserved for implementations.

CMakeMain.txt
yslking: Good idea. And Gtk also works on macOS, though cocoa works well.

CMakeLists.txt
yslking: ok.

docs/setup.md
```diff
@@ -170,6 +171,66 @@ If successful, cmake will generate a Visual Studio solution in the build
...
+```
+
+Note that GCC on MinGW may come across a multiple reference bug if you turn on LTO.
+To disable LTO, you should change
+```
+INTERPROCEDURAL_OPTIMIZATION TRUE
+```
+in `CMakeMain.txt` and `lib/CMakeLists.txt` to
+```
+INTERPROCEDURAL_OPTIMIZATION FALSE
```
djowel: Shouldn't we make this automatic if the mingw is detected?

docs/setup.md
```diff
@@ -4,6 +4,7 @@
 * [Requirements](#requirements)
 * [MacOS Installation](#macos)
 * [Windows Installation](#windows)
+* [Windows Installation (Gtk)](#windows-gtk)
```
djowel: Perhaps [Windows Installation (MinGW Gtk)] ?

docs/setup.md
Xeverous: I prefer a CMake option like `ELEMENTS_ENABLE_LTO`.

Xeverous:
@yslking let me know when the requested PR changes are made, I want to test the build.

yslking:
1. Change the CMake variable from `WINDOWS_GRAPHICS_LIBRARY` to `HOST_UI_LIBRARY`.
1. Add CMake code to check whether HOST_UI_LIBRARY is valid. (e.g. `foo` is not a valid value, and `cocoa` is not a valid value on windows).
2. Add macro `_HOST_UI_LIBRARY_GTK_`, `_HOST_UI_LIBRARY_COCOA_` or `_HOST_UI_LIBRARY_WIN32_` accordingly.
3. `WIN32` and `MACOS_BUNDLE` for `add_excutable` should be determined by the operating system instead of the UI library. I fixed my mistake on it.
4. Move `/utf-8` option for `msvc` to `lib/CMakeLists.txt` and `CMakeMain.txt`, because there has been some code about compile options in `lib/CMakeLists.txt`. I think putting them together makes things look clearer. 
```
target_compile_options(elements PRIVATE
    $<$<CXX_COMPILER_ID:GNU>:-Wall -Wextra -Wpedantic -ftemplate-backtrace-limit=0>
    $<$<CXX_COMPILER_ID:Clang>:-Wall -Wpedantic -ftemplate-backtrace-limit=0>
    $<$<CXX_COMPILER_ID:MSVC>:/W4 /wd4244 /wd4305 /wd4996 /wd4267 /wd4018>
)
```
5. Update document.
    1. change the title of mingw + gtk choice
    2. add a msvc + nmake guide. 
6. Keep win32 as default UI library for windows, to provide a consistent interface (Win32 native is used before my modification). 

> @yslking let me know when the requested PR changes are made, I want to test the build.
@Xeverous .

docs/setup.md
yslking: fixed.
I did this because I thought MinGW will fix this bug and push the fixed version to rolling-versioned downstream software (e.g. Arch, openSUSE Tumbleweed and msys2) within a few weeks, which seems to be a wrong idea now.

lib/include/elements/app.hpp
```diff
@@ -9,7 +9,7 @@
 #include <string>
 #include <infra/support.hpp>
 
-#if defined(__linux__)
+#if defined(_HOST_UI_LIBRARY_GTK_)
```
djowel: Oh, sorry, a single underscore + caps is also illegal: https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

"All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use."

djowel: I suggest something like ELEMENTS_HOST_UI_LIBRARY_GTK

djowel:
This is looking good, @yslking. Apart from the underscore (see my review comment), I think this is OK.

lib/include/elements/app.hpp
yslking: Sorry about that.

djowel:
To be on the safe side, I merged to develop first. @Xeverous might have some more comments, esp. on CMake.

PR #79, opened 2020-03-29
Replace boost::filesystem with std::filesystem

dmacvicar:
(Associated PR https://github.com/cycfi/json/pull/2 )

This PR replaces the dependency on _boost::filesystem_ with _std::filesystem_, allowing to rely only on the header parts of boost.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,128 @@
+name: Build
+on: [push, pull_request]
+env:
+  BUILD_TYPE: Release
+jobs:
+  build:
+    name: ${{ matrix.config.name }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+        - name: "Windows Latest MSVC"
+          os: windows-latest
+          cc: "cl"
+          cxx: "cl"
+          environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars32.bat"
+        - name: "Ubuntu Latest GCC"
+          os: ubuntu-latest
+          cc: "gcc"
+          cxx: "g++"
+          # If this env variable is present, we set BOOST_ROOT to its content
+          # Github workers put newer Boost versions there
+          boost_root: "BOOST_ROOT_1_72_0"
+        - name: "macOS Latest Clang"
+          os: macos-latest
+          cc: "clang"
+          cxx: "clang++"
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        path: elements
+    - name: Checkout infra repo
+      uses: actions/checkout@v2
+      with:
+        repository: cycfi/infra
+        path: infra
+    - name: Checkout json repo
+      uses: actions/checkout@v2
+      with:
+        repository: cycfi/json
+        path: json
+    - name: Install package dependencies
+      id: cmake_and_ninja
+      shell: cmake -P {0}
+      run: |
+        if ("${{ runner.os }}" STREQUAL "Windows")
+          execute_process(COMMAND choco install ninja)
+        elseif ("${{ runner.os }}" STREQUAL "Linux")
+          execute_process(COMMAND sudo apt-get install libcairo2-dev)
+          execute_process(COMMAND sudo apt-get install libgtk-3-dev)
+          execute_process(COMMAND sudo apt-get install ninja-build)
+        elseif ("${{ runner.os }}" STREQUAL "macOS")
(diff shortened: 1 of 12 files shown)


djowel:
I have some comments in https://github.com/cycfi/json/pull/2. I think this should be applied on an case-by-case basis using #ifdefs . Some compilers versions do not have std::filesystem in std.

djowel:
Oh and glad to see CI working! :-)

djowel:
BTW, moved target to develop

djowel:
BTW, this: https://github.com/gulrak/filesystem might be our way to remove boost dependency altogether. It's header only "without other C++17 dependencies". I'm going to try it out with the "artist" library currently WIP.

djowel:
> BTW, this: https://github.com/gulrak/filesystem might be our way to remove boost dependency altogether. It's header only "without other C++17 dependencies". I'm going to try it out with the "artist" library currently WIP.

OK, it works like a charm. While I'm at it, in the "artist" lib, also removed the need to clone infra. It looks really nice. I think it makes sense to do the same in the elements lib.

dmacvicar:
> > BTW, this: https://github.com/gulrak/filesystem might be our way to remove boost dependency altogether. It's header only "without other C++17 dependencies". I'm going to try it out with the "artist" library currently WIP.
> 
> OK, it works like a charm. While I'm at it, in the "artist" lib, also removed the need to clone infra. It looks really nice. I think it makes sense to do the same in the elements lib.

So that library acts as a compat library for older compilers, still being header only?
That sounds like a good solution. It would still use std::filesystem on capable compilers right?

Let me know what I can try to do around this. I can find some time one evening or weekends.

djowel:
> So that library acts as a compat library for older compilers, still being header only?
> That sounds like a good solution. It would still use std::filesystem on capable compilers right?
> 
> Let me know what I can try to do around this. I can find some time one evening or weekends.

Precisely. Here's what I did with artist: https://github.com/cycfi/artist/blob/windows_port/lib/include/artist/detail/filesystem.hpp

ghc::filesystem is in its own submodule, and although it has cmake, I never bothered to add it with add_subdirectory. I simply added its include directory.

Oh I did the same for infra (make it a submodule and have its header path via cmake target_include_directories). Here's the header structure:

https://github.com/cycfi/artist/tree/windows_port/lib

Xeverous:
> While I'm at it, in the "artist" lib, also removed the need to clone infra. It looks really nice. I think it makes sense to do the same in the elements lib.

So, changing infra's `clamp/clamp_min/clamp_max` to `std::clamp/min/max`? :)

djowel:
> > While I'm at it, in the "artist" lib, also removed the need to clone infra. It looks really nice. I think it makes sense to do the same in the elements lib.
> 
> So, changing infra's `clamp/clamp_min/clamp_max` to `std::clamp/min/max`? :)

I think you misunderstood what I said. No, that is not the only use of infra. Infra stays, but as a submodule.

djowel:
> > > While I'm at it, in the "artist" lib, also removed the need to clone infra. It looks really nice. I think it makes sense to do the same in the elements lib.
> > 
> > 
> > So, changing infra's `clamp/clamp_min/clamp_max` to `std::clamp/min/max`? :)
> 
> I think you misunderstood what I said. No, that is not the only use of infra. Infra stays, but as a submodule.

OK, so I did an investigation and indeed there's very little use of infra in elements (unlike some of the other libraries (e.g. Q audio dsp lib and QPlug audio plugin lib.)). The only uses are the clamps, `noncopyable` and `CYCFI_ASSERT`. So after thinking about it, reluctantly I'd say go for it. Let us remove the infra dependency:

`clamp/clamp_min/clamp_max` --> `std::clamp/max/min`
`CYCFI_ASSERT(x, msg)` --> assert(x && msg)
`noncopyable` --> c++11 ctor/assign `delete`

I said reluctantly because I am afraid of the proliferation of small common utilities later. For example, just yesterday, I added the template deleter function class in infra because it's becoming common to use static  custom `unique_ptr` deleters. On the other hand, I'd say I'll deal with that when it becomes too prevalent, like in the other libraries I have. Not now.

So if anyone wants to take on this, please do. I'll start with `artist`. @Xeverous and @dmacvicar , your intuitions are spot on!

dmacvicar:
https://github.com/cycfi/elements/pull/81 for the filesystem part. If that one is better, this one can be closed, and the discussion review moved there.

I will see how difficult is to replace the other parts you mention then.

djowel:
@dmacvicar What's up with the C.I. failures?

dmacvicar:
I will investigate the CI failures in the develop branch. (https://github.com/cycfi/elements/pull/82)

This PR had dependencies in PR's to other git repositories, therefore its CI is not very relevant.

PR #82, opened 2020-04-05
Merge master to get C.I. in develop branch

dmacvicar:
Merge the CI support into the develop branch.

Perhaps we even see some compile errors from the latest changes in this very same PR CI run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         path: elements
+        submodules: 'recursive'
     - name: Checkout infra repo
       uses: actions/checkout@v2
       with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,8 +17,8 @@ jobs:
           environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars32.bat"
         - name: "Ubuntu Latest GCC"
           os: ubuntu-latest
-          cc: "gcc"
-          cxx: "g++"
+          cc: "gcc-9"
+          cxx: "g++-9"
           # If this env variable is present, we set BOOST_ROOT to its content
           # Github workers put newer Boost versions there
           boost_root: "BOOST_ROOT_1_72_0"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,128 @@
+name: Build
+on: [push, pull_request]
+env:
+  BUILD_TYPE: Release
+jobs:
+  build:
+    name: ${{ matrix.config.name }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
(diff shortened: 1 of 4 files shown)


dmacvicar:
Ok. One error is:

```
../lib/include/elements/detail/filesystem.hpp:27:11: fatal error: 'ghc/filesystem.hpp' file not found
# include <ghc/filesystem.hpp>
          ^~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Which means the CI is probably not checking out the submodules. Will investigate.

However, I am puzzled why in any of platforms it is trying to build with the replacement library.

On Windows I am also seeing:

```
..\lib\include\elements/view.hpp(17): fatal error C1083: Cannot open include file: 'boost/asio.hpp': No such file or directory
```

dmacvicar:
>However, I am puzzled why in any of platforms it is trying to build with the replacement library.

```
The CXX compiler identification is GNU 7.5.0
```

This explains a lot. The _Github_ runner preinstalls lot of "newer" software, but we need to point _cmake_ to it, as by default, _Ubuntu_ ships with very ancient stuff, and locally I am using _openSUSE Tumbleweed_ which has very recent everything.

dmacvicar:
@djowel  Ok. The errors we have left is because the json library still depends on boost filesystem.

djowel:
> @djowel Ok. The errors we have left is because the json library still depends on boost filesystem.

I see. How do you propose we proceed? Are you going to have a PR for that as well so it can propagate outwards? I can do that of course, but I am in the middle of something, so it might take time.

djowel:
Hmm, how about we move the filesystem workaround to infra? That way json, artist and elements can just use it. We can move infra in a submodule as well, like I did in artist, referring to it by include file only. Thoughts?

djowel:
Meantime, I'll fix develop.

djowel:
> Hmm, how about we move the filesystem workaround to infra? That way json, artist and elements can just use it. We can move infra in a submodule as well, like I did in artist, referring to it by include file only. Thoughts?

OK, I am doing it anyway, starting from infra then json. I think it makes sense to have these as submodules. I'll need it to fix develop. Let's see how it goes...

djowel:
Develop should be good now. See 115dca5dfbda41f81bc26d72f9c1561cb01fb6a6.

dmacvicar:
> > @djowel Ok. The errors we have left is because the json library still depends on boost filesystem.
> 
> I see. How do you propose we proceed? Are you going to have a PR for that as well so it can propagate outwards? I can do that of course, but I am in the middle of something, so it might take time.

I will submit a PR that uses std:: or ghc:: later. I will make sure it uses the ghc from infra.

djowel:
Oh, and I think we can remove the need to checkout infra and json from the C.I. scripts now. I also adjusted the cmake scripts. infra and json are now just submodules.

djowel:
Hi @dmacvicar 

For some reason the Windows build could not find boost in this commit:
https://github.com/cycfi/elements/runs/608726683?check_suite_focus=true

Do you have an idea what's going on? There were no changes to cmake in that commit, just code changes. The commit before that is fine.

BTW, is there a way to contact you vis email, private message or something else?

PR #83, opened 2020-04-06
CI: No need to checkout infra and json separately

dmacvicar:
Now that infra and json are submodules, we only need to checkout the main elements repo, and do not need to build inside a directory of the workspace.


djowel:
I'll need to update the Setup and Installation docs as well.

PR #85, opened 2020-04-06
docs: Update build instructions

dmacvicar:
Reflect that now infra and json are submodules.

Perhaps Boost instructions should be changed as well, but it is not part of this PR.


djowel:
Nice! You beat me to it :-)

PR #90, opened 2020-04-26
move semantics improvements

Xeverous:


djowel:
There's currently a problem with C.I. on windows. It can't seem to find Boost. I haven't looked at it yet. I'll probably need to get familiar with the C.I. mechanism.

PR #99, opened 2020-05-02
make LTO an explicit CMake option

Xeverous:


Xeverous:
LTO with MinGW is more or less broken, but this commit is much better than any assumptions in the build recipe.

Xeverous:
Amazing. Please read GCC build logs. You will be surprised, but this is actually a good thing for my case.

djowel:
lto1: internal compiler error: in add_symbol_to_partition_1, at lto/lto-partition.c:153
Please submit a full bug report,

Xeverous:
So what do we do with this now? IMO now it is even clearer that LTO should be an explicit option. Perhaps make it OFF by default?

djowel:
That makes sense. But it seems clang is good, correct? So maybe on by default on clang?

Xeverous:
I don't think so. Better be consistent and do not assume anything. IMO LTO should be an explicit, independent option just like `CMAKE_BUILD_TYPE`.

djowel:
> I don't think so. Better be consistent and do not assume anything. IMO LTO should be an explicit, independent option just like `CMAKE_BUILD_TYPE`.

OK, you are right.

Xeverous:
Changed default to OFF, now CI fully succeeds.

djowel:
Many thanks, @Xeverous !

Xeverous:
> Many thanks, @Xeverous !

I will probably work on modals today. Can you look into my tab prototype commit? Notebook is the most important thing for me now.

PR #102, opened 2020-05-04
fix doc link and add missing bracket

pkoretic:
Hi!
Just noticed this while going through the docs, although I understand you want to do more work on the docs but this won't hurt in the meantime.


djowel:
Wonderful! Thanks! yeah, working on more docs now. You might want to check out:
https://cycfi.github.io/elements/layout

PR #114, opened 2020-05-26
Remove the dependency Boost.Asio

jpcima:


jpcima:
Sorry didn't mean to target this PR to here.. it was aimed at personal repo, please ignore.
Sorry for noise

djowel:
It's interesting though. How are you handling async tasks though? There are things like view.refresh() that needs. to run in the main thread, but can be called in background threads.

jpcima:
>It's interesting though. How are you handling async tasks though?

This implementation is not thread-safe at all.

>can be called in background threads.

How does the background call happen? is that specific to some particular UI hosts?

For context, the use case of this fork is use in audio plugin, which is for all systems including Linux. A part of this is Pugl as you know, but also I've wanted to backport some parts to C++14 to meet some constraints.
(and as well I'd like to avoid some dependencies to the possible extent, but that is not as much a hard requirement)

ps. thanks a lot for creating this library, and appreciate your help !

djowel:
> How does the background call happen? is that specific to some particular UI hosts?
> 
> For context, the use case of this fork is use in audio plugin, which is for all systems including Linux. A part of this is Pugl as you know, but also I've wanted to backport some parts to C++14 to meet some constraints.
> (and as well I'd like to avoid some dependencies to the possible extent, but that is not as much a hard requirement)
> 
> ps. thanks a lot for creating this library, and appreciate your help !

When your code wants to update the state of the UI (e.g. VST3 or AU parameter changes, e.g. slider values changes that follow the track recording changes or perhaps MIDI changes), there is no guarantee that such parameter changes happen in the main thread. So, the update mechanism should be thread safe. In certain OSes (such as MacOS), drawing can only happen in the main thread and only within drawing events, so the way to do a UI update is via the refresh mechanism, but refresh also has to happen in the main thread or will be ignored (e.g. https://github.com/glfw/glfw/issues/136).

Issue #123, opened 2020-05-29
Fonts in custom resource path on Linux, cannot load


jpcima:
Hello, on Linux it's impossible to load the font shipped along with my plugin.
The library cannot deduce a resource path on this OS, so it decides one based on the work directory.
This technique is unable to work in particular cases like audio plugins.

https://github.com/cycfi/elements/blob/e29d1fad659291a25294f8d93798ef223ef7dc4f/lib/src/support/font.cpp#L156

What is the preferred solution to resolve this problem?
I would be satisfied if there exists a way to set up the font path externally, similar to the resource path.

djowel:
Looks like you hit a bug. The right way to deal with this is to provide a generic to way to allow users to set fonts path or paths, perhaps similar to the way resources such as images are loaded. This should be easy. I'll think of a way, but feel free to suggest!

jpcima:
The solution has worked, I close it.

PR #124, opened 2020-05-29
Allow to set custom paths for fonts

jpcima:
#123

This adds a variable `font_paths` for setting custom paths.
It has to be set early, before the program initializes the font map.

A problem is that the global theme is statically initialized, and so is the font map as a consequence.
I replace the global theme static in order to make it lazily instantiated.

### DIFF
--- a/lib/include/elements/support/font.hpp
+++ b/lib/include/elements/support/font.hpp
@@ -8,6 +8,7 @@
 
 #include <string_view>
 #include <infra/filesystem.hpp>
+#include <vector>
 
 extern "C"
 {
@@ -308,6 +309,8 @@ namespace cycfi { namespace elements
 #ifdef __APPLE__
    fs::path get_user_fonts_directory();
 #endif
+
+   extern std::vector<fs::path> font_paths;
 }}
 
 #endif
--- a/lib/include/elements/support/font.hpp
+++ b/lib/include/elements/support/font.hpp
@@ -310,7 +310,7 @@ namespace cycfi { namespace elements
    fs::path get_user_fonts_directory();
 #endif
 
-   extern std::vector<fs::path> font_paths;
+   std::vector<fs::path>& font_paths();
 }}
 
 #endif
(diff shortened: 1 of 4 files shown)


lib/include/elements/support/font.hpp
```diff
@@ -308,6 +309,8 @@ namespace cycfi { namespace elements
 #ifdef __APPLE__
    fs::path get_user_fonts_directory();
 #endif
+
+   extern std::vector<fs::path> font_paths;
```
djowel, line 313: I think it's better to make this a function that returns a "lazily" instantiated font_paths to avoid potential static initialization order issues.

lib/src/support/font.cpp
```diff
@@ -150,18 +150,22 @@ namespace cycfi { namespace elements
...
          TCHAR windir[MAX_PATH];
          GetWindowsDirectory(windir, MAX_PATH);
-         auto fonts_path = (fs::path(windir) / "fonts").generic_string();
-         FcConfigAppFontAddDir(config, (FcChar8 const*)fonts_path.c_str());
+         paths.push_back(fs::path(windir) / "fonts");
 #endif
 #endif
-         FcConfigAppFontAddDir(config, (FcChar8 const*)resources_path.c_str());
+         for (auto& path : paths)
+            FcConfigAppFontAddDir(config, (FcChar8*)path.generic_string().c_str());
```
djowel, line 167: Doesn't want the `const` ? It's probably good practice to use c++ casting style (I know I sometimes forget :-) ).

Xeverous: Yes, I would definitely recommend named casts. It's not about shorter code, it's about clear code for the reader. Also, `-Wold-style-cast` and Core Guidelines...

lib/include/elements/support/font.hpp
jpcima: Agreed, do you think this solution to be appropriate?
46596836088e67c7dc23860a15a7c1bf92492a1c

djowel: Looks good to me

Issue #88, opened 2020-04-14
Remove string duplicates in widget factories


Xeverous:
There are multiple factories like this one which unnecesarily copy the string:

https://github.com/cycfi/elements/blob/ccb2f2c4107827cb9f9d98e8fb3daf3944e0567f/lib/include/elements/element/gallery/check_box.hpp#L35-L41

Instead of creating a such hierarchy:

- check_box
  - htile
    - layered_element
      - icon_on
      - icon_off
    - text

they create this:

- check_box
  - layered_element
    - htile
      - icon_on
      - text
    - htile
      - icon_off
      - text
___

This issue is blocked by #87

Xeverous:
I see a bigger problem there: many factories are duplicating elements or their data in order to create 2 elements that represent some boolean state. I think this is a good place for another element type that will abstract such state. Types such as `layered_button` and `basic_choice`, for their ctors instead of taking 2 elements could then take such single element. This could offer to not copy the stored element if both states are the same while also having a way to actually have 2 separate elements.

Xeverous:
All such factories:

```
$ grep -rn -C 7 "<false" lib/include/
lib/include/elements/element/gallery/check_box.hpp-31-   {
lib/include/elements/element/gallery/check_box.hpp-32-      draw_check_box(ctx, _text, state, ctx.bounds.includes(ctx.view.cursor_pos()));
lib/include/elements/element/gallery/check_box.hpp-33-   }
lib/include/elements/element/gallery/check_box.hpp-34-
lib/include/elements/element/gallery/check_box.hpp-35-   inline basic_toggle_button<> check_box(std::string text)
lib/include/elements/element/gallery/check_box.hpp-36-   {
lib/include/elements/element/gallery/check_box.hpp-37-      return basic_toggle_button(
lib/include/elements/element/gallery/check_box.hpp:38:         check_box_element<false>{ text }
lib/include/elements/element/gallery/check_box.hpp-39-       , check_box_element<true>{ text }
lib/include/elements/element/gallery/check_box.hpp-40-      );
lib/include/elements/element/gallery/check_box.hpp-41-   }
lib/include/elements/element/gallery/check_box.hpp-42-}}
lib/include/elements/element/gallery/check_box.hpp-43-
lib/include/elements/element/gallery/check_box.hpp-44-#endif
--
lib/include/elements/element/gallery/icon_button.hpp-63-   inline basic_toggle_button<> toggle_icon_button(
[...]

PR #128, opened 2020-05-31
tab factory: avoid copying elements

Xeverous:


djowel:
I'll merge this for now, but so you know, I've been thinking about the ON-OFF duplicated stuff issue and I might be doing something in that end.

Xeverous:
Great, see my comment in #88 then.

I made this change for 2 reasons:

- replace the copy with 2 explicitly given elements to avoid copying problem
- make the tab factory to be consistent with other factories that create a composite of 2 elements - this is another problem, but after this PR thanks to consistency the final resolution for on/off problem will be the same for every such factory

Issue #130, opened 2020-06-02
Offline documentation using Jekyll locally


redtide:
I was playing with the website pages to load them locally but I'm not sure if it could be a useful feature for you; I saw recently a commit on master that you use to link to website pages instead local ones plus, AFAIK, Jekyll on Windows is problematic (no idea under macOS, should be fine I suppose), otherwise I could send a PR for this.

djowel:
> I was playing with the website pages to load them locally but I'm not sure if it could be a useful feature for you; I saw recently a commit on master that you use to link to website pages instead local ones plus, AFAIK, Jekyll on Windows is problematic (no idea under macOS, should be fine I suppose), otherwise I could send a PR for this.

Sure, go ahead! Sounds good.

redtide:
Thanks! Closing.

PR #131, opened 2020-06-02
Adapt docs files to be viewed locally

redtide:
I haven't a way to test this online since GH permits pages only from master branch, locally it's correct, I'm not sure about the relative links if they will work correctly as is or they needs a `| relative_url` addition later.
I've added also a simple README file, some more .gitignored files (VSCode works good with Jekyll) and a convenient bash script to install and run things under a Linux environment (locally at least here on ArchLinux using RVM).
The added Gemfile is required to install bundler and other dependencies required to run Jekyll.
See #130.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,6 +1,7 @@
 _site
+.bundle
 .sass-cache
 .jekyll-metadata
 /Gemfile.lock
-/Gemfile
 /404.html
+*.code-workspace
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,0 +1,21 @@
+source "https://rubygems.org"
+
+# Force gh-pages compatibility in RVM without gh-pages installed
+# see https://pages.github.com/versions/
+ruby "2.5.3"
+
+# Uncomment to build our site in Travis-CI
+# see https://docs.travis-ci.com/user/languages/ruby/#default-build-script
+# and https://github.com/travis-ci/travis-web/blob/master/Gemfile
+#group :development, :test do
+#  gem "rake", "~> 12"
+#end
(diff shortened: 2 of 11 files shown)


djowel:
OK, this works. verified... I had to change ruby to "2.6.0" though, because that is what I do locally when working on the docs. Would you have a problem with that?

djowel:
Update: actually wouldn't it be better if we don't specify ruby "2.6.0"? That is what I originally have in my local setup.

djowel:
Update: 

- I had to fix the README.md to. The website will be available at <http://localhost:4000/elements/>.
- I had to fix jeckyl.sh, removing --host=0.0.0.0 for the README.md to be correct. Is there a problem on Windows? Why did you have --host=0.0.0.0?

redtide:
> actually wouldn't it be better if we don't specify ruby "2.6.0"? That is what I originally have in my local setup.

As specified in the README, I just try to keep the same versions as used by GitHub. This to avoid conflicts, so for that reason I've installed RVM to keep all the required Ruby versions. I can just imagine that some day the user that needs to run the pages could have a new version that will not work (well) with the old Jekyll used there or try to update Jekyll and have troubles because deprecated APIs. I think that happened to me at least 2 times, so I have become paranoid because of that :-D

> * I had to fix the README.md to. The website will be available at http://localhost:4000/elements/.

Sorry, I forgot about the baseurl.

> * I had to fix jeckyl.sh, removing --host=0.0.0.0 for the README.md to be correct. Is there a problem on Windows? Why did you have --host=0.0.0.0?

That should make the website accessible also from other machines in the LAN.

PR #136, opened 2020-06-06
reimplementation of tile space algorithm redistribution

Xeverous:
replaced existing algorithm which was not correct (sort of brute
force calculation with invalid loop condition, division by zero,
float/double inconsistency) with a new one that has O(n log n)
complexity and floating-point related checks

also, reduced some unnecessary lifetime

### DIFF
--- a/lib/src/element/tile.cpp
+++ b/lib/src/element/tile.cpp
@@ -1,10 +1,14 @@
 /*=============================================================================
-   Copyright (c) 2016-2020 Joel de Guzman
+   Copyright (c) 2016-2020 Joel de Guzman, Michal Urbanski
 
    Distributed under the MIT License [ https://opensource.org/licenses/MIT ]
 =============================================================================*/
 #include <elements/element/tile.hpp>
 #include <elements/support/context.hpp>
+
+#include <algorithm>
+#include <numeric>
+#include <limits>
 
 namespace cycfi { namespace elements
 {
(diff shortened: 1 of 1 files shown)


djowel:
@Xeverous, this code is fantastic! This is the best implementation so far in the evolution of the tiling code. The behavior of which has only been formalized recently, which was captured to the layout documentation I wrote recently. The tiling algorithm is one of the most important aspects of Elements and there's been quite a few quirks here and there until the formalization of its behavior.

Thank you very much!!!

Xeverous:
When I first read the docs and saw the stretch, I immediately thought it to be equivalent to [`flex-grow`](https://cssreference.io/property/flex-grow/) CSS property. The only difference is that in elements library each element can also have a max size, so simple `stretch / sum_stretch * free_space` is not enough because some elements will reach their max size and there will be still some free space to redistribute.

The previous implementation did exactly that simple formula while looping over elements and then looped again to redistribute remaining space. If the loop condition was correct (untill no space is left or all elements reach max instead of hardcoded to loop 9 times) it could be a valid implementation but such one has quadratic complexity.

Initially I wanted to fix the dvision by zero and other floating-point errors but when I noticed it is quadratic with no correct loop condition I went for a full rewrite. I was pretty sure it can be done in linear or log-linear time (if sorting was needed). Only thing left was to figure out [how to loop just once by updating sum of stretch weights on the go](https://www.reddit.com/r/algorithms/comments/gx6qbo/looking_for_algorithm_name_or_implementation_for/).

I'm a bit interested in algorithms and may at some point do [stuff like this](https://www.youtube.com/watch?v=kPRA0W1kECg) (obviously using elements).

djowel:
> Initially I wanted to fix the dvision by zero and other floating-point errors but when I noticed it is quadratic with no correct loop condition I went for a full rewrite. I was pretty sure it can be done in linear or log-linear time (if sorting was needed). Only thing left was to figure out [how to loop just once by updating sum of stretch weights on the go](https://www.reddit.com/r/algorithms/comments/gx6qbo/looking_for_algorithm_name_or_implementation_for/).

Way to go, @Xeverous. I was going to tackle this somehow as a future TODO (yes, I am arware of the big-O complexity). There was just so many things to do that I tend to always optimize later. Again, thank you for the splendid implementation!

> Only thing left was to figure out [how to loop just once by updating sum of stretch weights on the go](https://www.reddit.com/r/algorithms/comments/gx6qbo/looking_for_algorithm_name_or_implementation_for/).

Cool! You should document that implementation in the layout docs! :-) It would be nice to have it featured as part of the docs.

PR #138, opened 2020-06-08
RAII for FreeType and FontConfig libraries

Xeverous:
this fixes a lot of leaks

### DIFF
--- a/lib/src/support/font.cpp
+++ b/lib/src/support/font.cpp
@@ -1,5 +1,5 @@
 /*=============================================================================
-   Copyright (c) 2016-2020 Joel de Guzman
+   Copyright (c) 2016-2020 Joel de Guzman, Michal Urbanski
 
    Distributed under the MIT License [ https://opensource.org/licenses/MIT ]
 =============================================================================*/
@@ -31,10 +31,13 @@
 
 #include <map>
 #include <mutex>
+#include <memory>
 #include <sstream>
 #include <algorithm>
 #include <vector>
 #include <utility>
+#include <optional>
+#include <type_traits>
 
 namespace cycfi { namespace elements
 {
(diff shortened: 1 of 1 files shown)


Xeverous:
Now this is ready.

djowel:
I think this PR is "fixing" way more than necessary. I'm not sure I like it. I'll have a go with asan, (which I should have done already, regrettably!) and I'll see if I can fix with minimal changes.

Xeverous:
This is minimal set of changes. A class wrapping the resource (manually written or `std::unique_ptr` with custom deleter) is the only way to ensure resource safety - resources allocated in a different place than constructor+destructor pair are not guuaranteed to work. There is a lot of new code because the old code did not clean up the resources at all.

djowel:
> This is minimal set of changes. A class wrapping the resource (manually written or `std::unique_ptr` with custom deleter) is the only way to ensure resource safety - resources allocated in a different place than constructor+destructor pair are not guuaranteed to work. There is a lot of new code because the old code did not clean up the resources at all.

I am not sure I agree.

djowel:
Anyway, as I said, I think I better get working with the sanitizers in my dev setup. I'll see if I can get something simpler going and assess whether the additional code you propose is actually needed.

Xeverous:
> in many cases RAII is not even needed.

True, we can always code `if (err) goto cleanup;` if you like it...

Xeverous:
> I better get working with the sanitizers in my dev setup.

This works for me:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 1aaa336..91d1c3e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,9 @@ option(ELEMENTS_BUILD_EXAMPLES "build Elements library examples" ON)
 option(ELEMENTS_ENABLE_LTO "enable link time optimization for Elements targets" OFF)
 set(ELEMENTS_HOST_UI_LIBRARY "" CACHE STRING "gtk, cocoa or win32")
 option(ELEMENTS_HOST_ONLY_WIN7 "If host UI library is win32, reduce elements features to support Windows 7" OFF)
+option(ELEMENTS_SANITIZE_ADDRESS "Enable address sanitizer" OFF)
+option(ELEMENTS_SANITIZE_UNDEFINED "Enable undefined behavior sanitizer" OFF)
+option(ELEMENTS_SANITIZE_THREAD "Enable thread sanitizer" OFF)

 add_subdirectory(lib)

diff --git a/CMakeMain.txt b/CMakeMain.txt
index c27ecac..697e0ea 100644
--- a/CMakeMain.txt
+++ b/CMakeMain.txt
@@ -129,6 +129,21 @@ target_compile_options(${ELEMENTS_APP_PROJECT} PRIVATE
     $<$<CXX_COMPILER_ID:MSVC>:/utf-8>
 )

+###############################################################################
+# sanitizers
+
+if (ELEMENTS_SANITIZE_ADDRESS)
+   target_compile_options(${ELEMENTS_APP_PROJECT} PRIVATE -fsanitize=address)
+endif()
+
+if (ELEMENTS_SANITIZE_UNDEFINED)
+   target_compile_options(${ELEMENTS_APP_PROJECT} PRIVATE -fsanitize=undefined)
+endif()
+
+if (ELEMENTS_SANITIZE_THREAD)
+   target_compile_options(${ELEMENTS_APP_PROJECT} PRIVATE -fsanitize=thread)
+endif()
+
 ###############################################################################
 # Libraries and linking

diff --git a/lib/CMakeLists.txt b/lib/CMakeLists.txt
index 4cfff85..5f2743c 100755
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -165,6 +165,30 @@ target_sources(elements
 target_include_directories(elements PUBLIC include)

 target_compile_features(elements PUBLIC cxx_std_17)
+
+###############################################################################
+# sanitizers
+
+if (ELEMENTS_SANITIZE_ADDRESS)
+   message(STATUS "Compiling elements with address sanitizer")
+   target_compile_options(elements PRIVATE -fsanitize=address)
+   target_link_libraries(elements PUBLIC asan)
+endif()
+
+if (ELEMENTS_SANITIZE_UNDEFINED)
+   message(STATUS "Compiling elements with undefined behavior sanitizer")
+   target_compile_options(elements PRIVATE -fsanitize=undefined)
+   target_link_libraries(elements PUBLIC ubsan)
+endif()
+
+if (ELEMENTS_SANITIZE_THREAD)
+   message(STATUS "Compiling elements with thread sanitizer")
+   target_compile_options(elements PRIVATE -fsanitize=thread)
+   target_link_libraries(elements PUBLIC tsan)
+endif()
+
+###############################################################################
+
 target_link_libraries(elements PUBLIC cycfi::infra cycfi::json)

 if(ELEMENTS_ENABLE_LTO)
```

Note that sanitizers need to be first in linked libraries because they replace some C and C++ standard library functions.

Xeverous:
> assess whether the additional code you propose is actually needed

Which code exactly you consider additional?

djowel:
> True, we can always code `if (err) goto cleanup;` if you like it...

Do you enjoy sarcasms? ;-) It won't help, IMO.

djowel:
> > I better get working with the sanitizers in my dev setup.
> 
> This works for me:

Would you mind having that merged in a PR? This will go well with the examples and the tests I am working on now.

Xeverous:
> Do you enjoy sarcasms? ;-) It won't help, IMO.

Sometimes, yes (this sentence is not a sarcasm).

Xeverous:
> Do you enjoy sarcasms? ;-) It won't help, IMO.

This is a bit more complex that my diff. This diff works for me (GCC 8 on 64-bit GNU/Linux) but it is case-specific.

`-fsanitize` needs to be both in compiler and linker options and there may be more compiler-specific differences. The runtime needs also to be loaded first.

djowel:
I merged this already. See my comments on https://github.com/cycfi/elements/pull/139

Further, for future improvements: It is possible to capture the essence of the RAII involved in this PR in an idiomatic, generic template helper that can be placed in support.

Xeverous:
> It is possible to capture the essence of the RAII involved in this PR in an idiomatic, generic template helper that can be placed in support.

Isn't unique pointer such thing already? By supplying custom deleter you can easily make a non-copyable type that has RAII. I went for classes in some places because there were additional functions using the handle. How such template would look like?

djowel:
> > It is possible to capture the essence of the RAII involved in this PR in an idiomatic, generic template helper that can be placed in support.
> 
> Isn't unique pointer such thing already? By supplying custom deleter you can easily make a non-copyable type that has RAII. I went for classes in some places because there were additional functions using the handle. How such template would look like?

Perhaps you are right. I'm not sure. To be honest, I haven't had the chance to look at it closely yet.

Xeverous:
The core of the issue is to implement a rule-of-5 moveable or copyable type. Trivial non-moveable types are basically unique pointers which just need a custom factory (I just write some function that starts with `make_`) and the deleter. I don't think you can make it simpler or abstract more - the rules for RAII are the same everywhere, you only need to setup construction and destruction functions.

Such auto-wrapping could be potentially more automated, but it would either require a lot of preprocessor or proposed metaclasses feature.

Issue #141, opened 2020-06-10
Make {CMakeMain|CMakeInclude}.txt as CMake modules


redtide:
I've made [this attempt] to make more explicit what the 2 files are about, 1 as application build configuration and the other as a common one for both library and applications using it, as previously mentioned [in this issue]. I moved and renamed them in the usual CMake modules directory and added some minor missing definitions on some examples.
I'm not sure if it's the proper solution, in case it will be OK I'll post a pull request.

[this attempt]: https://github.com/redtide/elements/commit/4f2e4fb22566a498cfdf2009416390c6a29d757d
[in this issue]: https://github.com/cycfi/elements/issues/89#issuecomment-636753358

djowel:
I suppose this affects the setup and installation guide? 

"Building Outside the Project" in http://cycfi.github.io/elements/setup

redtide:
All I did is to rename 2 CMake files, moved them on a CMake module directory, add it to the CMake module path and adapt the application script to use ELEMENTS_ROOT instead CMAKE_CURRENT_LIST_DIR because I was getting some path error, otherwise it should work as it previously did.

djowel:
Sounds good to me.

PR #150, opened 2020-06-12
Move {CMakeMain|CMakeInclude}.txt as CMake modules

redtide:
I had to redo the original branch because it was branched from master instead develop by mistake.
I've added a ignored file for the IDE I use currently if that's OK.
See #141.


djowel:
> I've added a ignored file for the IDE I use currently if that's OK.

Should be fine.

djowel:
Many thanks, @redtide !!!

redtide:
Thanks to you for the library!

Issue #135, opened 2020-06-02
leaks in font map


Xeverous:
Welcome to another episode of *Why I hate global variables* although this time the problem is more due to a lack of RAII (nonetheless I would like to get rid of as many globals as possible, possibly moving them to instantiate by library users explicitly).

Whenever I close any of examples which has some text, I'm getting *SPAMMED* with this and similar:

```
Indirect leak of 192 byte(s) in 6 object(s) allocated from:
    #0 0x7fbc79fc7f90 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xedf90)
    #1 0x7fbc78f958ed in FcStrSetCreate (/usr/lib/x86_64-linux-gnu/libfontconfig.so.1+0x218ed)
    #2 0x55564e0f315b in match /home/xeverous/workspace/elements/lib/src/support/font.cpp:221
    #3 0x55564e0f4baa in cycfi::elements::font::font(cycfi::elements::font_descr) /home/xeverous/workspace/elements/lib/src/support/font.cpp:292
    #4 0x55564e04f292 in cycfi::elements::theme::theme() /home/xeverous/workspace/elements/lib/src/support/theme.cpp:71
    #5 0x55564e0511b7 in cycfi::elements::global_theme::_theme() /home/xeverous/workspace/elements/lib/src/support/theme.cpp:78
    #6 0x55564e05121a in cycfi::elements::get_theme() /home/xeverous/workspace/elements/lib/src/support/theme.cpp:84
    #7 0x55564dfe84d8 in cycfi::elements::default_label::font_size() const (/home/xeverous/workspace/elements/build/examples/doc_aspects/Aspects+0x3b94d8)
    #8 0x55564e007af0 in cycfi::elements::default_label::limits(cycfi::elements::basic_context const&) const /home/xeverous/workspace/elements/lib/src/element/label.cpp:16
    #9 0x55564dff2588 in cycfi::elements::margin_element<cycfi::elements::rect, cycfi::elements::label_gen<cycfi::elements::basic_label_base<cycfi::elements::default_label> > >::limits(cycfi::elements::basic_context const&) const (/home/xeverous/workspace/elements/build/examples/doc_aspects/Aspects+0x3c3588)
    #10 0x55564e00ab88 in cycfi::elements::layer_element::limits(cycfi::elements::basic_context const&) const /home/xeverous/workspace/elements/lib/src/element/layer.cpp:20
[...]

PR #157, opened 2020-06-15
Do not deinitialize the fontconfig library

jpcima:
#135

This skips deinitializing the fontconfig library.
It prevents crashes in a DAW which also use the system-wide fontconfig library.

- if the host application takes care of deinitializing fontconfig, it's fine and memory leaks should not occur.
- otherwise, it will be just a negligible leak of some hundreds kB, not cumulative when instances of Elements ui will be created and destroyed repeatedly.

### DIFF
--- a/lib/src/support/font.cpp
+++ b/lib/src/support/font.cpp
@@ -92,66 +92,38 @@ namespace cycfi { namespace elements
          struct load_none_tag {};
          struct no_init_tag {};
 
-         class library
+         class config
          {
          public:
-            library(no_init_tag)
+            config(no_init_tag)
             {
             }
 
-            library(load_none_tag)
-            : _config_ptr(nullptr)
-            , _is_initialized(FcInit() == FcTrue)
+            config(load_none_tag)
+            : _config_ptr((FcInit() == FcTrue) ? FcConfigCreate() : nullptr)
             {
             }
 
-            library(load_config_tag)
-            : _config_ptr(FcInitLoadConfig())
-            , _is_initialized(_config_ptr != nullptr)
+            config(load_config_tag)
+            : _config_ptr((FcInit() == FcTrue) ? FcInitLoadConfig() : nullptr)
             {
             }
 
-            library(load_config_and_fonts_tag)
-            : _config_ptr(FcInitLoadConfigAndFonts())
-            , _is_initialized(_config_ptr != nullptr)
+            config(load_config_and_fonts_tag)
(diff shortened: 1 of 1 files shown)


Xeverous:
I don't like such change:

- I would rather move all singletons upwards to the public API so they can be explictly instantiated (or not) by elements library users.
- The code after this commit leaves `git blame` in a state as if I wrote the RAII wrapper incorrectly. At this point the wrapper can be removed entirely as it does not implement what it was written for - RAII. This PR basically breaks a working class.

> it will be just a negligible leak of some hundreds kB, not cumulative when instances of Elements ui will be created and destroyed repeatedly

I dislike such argumentation. If had a project which used X-Y libraries, and if many or every of them went this way, I would be furious at all the sanitizers reports while trying to detect or fix bugs within my own code.

jpcima:
Well yeah but the deal here, it's that fontconfig is clearly not a library that was designed to accept RAII, it lives by a C-inspired design that it should be global, attached to the lifetime of the application.

It's this problem that we have to live with design decisions made by others, and there is need of a practical working solution.
Regarding the wrapper class, your argument makes sense that it would lose its RAII vocation that it has previously, fine to rewrite it in another way as needed.

Btw, regarding the basic program `FcInit(); FcFfini();`.
The sanitizer shows allocations anyway, because fontconfig authors probably have not cared enough to take care of them.
```
==123770== HEAP SUMMARY:
==123770==     in use at exit: 18,612 bytes in 6 blocks
==123770==   total heap usage: 10,559 allocs, 10,553 frees, 3,450,025 bytes allocated
```

jpcima:
New suggestion of fontconfig RAII wrapper. Only for `FcConfig`.

If the programmers wish to clean up the fontconfig global kilobytes, which I doubt it matters a lot, the program made with Elements can always be wrapped between `FcInit(); [...]; FcFini();`.

djowel:
> Well yeah but the deal here, it's that fontconfig is clearly not a library that was designed to accept RAII, it lives by a C-inspired design that it should be global, attached to the lifetime of the application.

I'm with @jpcima on this. It does not really hurt for some resources to be tied to the lifetime of the application.

djowel:
> New suggestion of fontconfig RAII wrapper. Only for `FcConfig`.
> 
> If the programmers wish to clean up the fontconfig global kilobytes, which I doubt it matters a lot, the program made with Elements can always be wrapped between `FcInit(); [...]; FcFini();`.

Please do. Anything that will prevent crashing a DAW is a good solution for me. A few global kilobytes does not matter!

PR #159, opened 2020-06-18
Fix a problem of freetype font lifetime

jpcima:
This seems to work around a crash, where the freetype font is destroyed as it's still in use.
The circumstances of the problem are not clear.

It would happen when opening/closing a view many times in a row in an audio plugin host.
(not any plugin host, it seems to happen in Ardour6 but not Carla for instance)

The best I have about it is a sequence of valgrind backtraces.
[vg-ardour6-100.txt.gz](https://github.com/cycfi/elements/files/4796383/vg-ardour6-100.txt.gz)

When it occurs, it seems that freetype returns an identical `ft_face` handle to one that was previously loaded.

The mishandling of references seems possibly happens under cairo, together with these duplicate references thing. Perhaps the better solution is not letting cairo to handle this reference counting (that this user data mechanism seems for), but do it in RAII at library side.

### DIFF
--- a/lib/src/support/font.cpp
+++ b/lib/src/support/font.cpp
@@ -515,7 +515,6 @@ namespace cycfi { namespace elements
             return _face;
          }
 
-         [[nodiscard]]
          FT_Face release()
          {
             return std::exchange(_face, nullptr);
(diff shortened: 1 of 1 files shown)


djowel:
Thank you @jpcima

PR #168, opened 2020-06-25
fixed utf-8 handling for text boxes

petiaccja:


petiaccja:
I was wondering, though, wouldn't it be easier to just store all in-memory strings as an std::u32string? More memory, yes, but then there are no problems with the quirks of multi-byte characters.

djowel:
> I was wondering, though, wouldn't it be easier to just store all in-memory strings as an std::u32string? More memory, yes, but then there are no problems with the quirks of multi-byte characters.

Gee, thanks! The artist port of the text layout management does indeed use u32 internally, most of the time. So perhaps that will be the likely outcome anyway. For now, let's keep it as it is.

PR #169, opened 2020-06-28
fixed u8 strings under c++20

petiaccja:
With this patch, *elements* compiles with C++20 mode (MSVC), but there isn't really an elegant way to fix this problem. Maybe something else works better for *elements*, meaning this PR may just be completely reworked or treated as a temporary fix.

What do you think about adding a C++20 build to the CI?

Reference: [http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r2.html](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r2.html)

### DIFF
--- a/lib/src/element/gallery/menu.cpp
+++ b/lib/src/element/gallery/menu.cpp
@@ -5,6 +5,19 @@
 =============================================================================*/
 #include <elements/element/gallery/menu.hpp>
 
+
+// Instead of
+// std::string s = u8"string",
+// write 
+// std::string s = u8_compat("string").
+//
+// In C++20, u8"strings" are not compatible with std::string anymore.
+// Practically, one must either use std::u8string (which is not available in C++17),
+// or force u8"strings" to keep creating const char*'s (even with C++20).
+// This macro does the latter.
+#define u8_compat(str) reinterpret_cast<const char*>(u8##str);
+
+
 namespace cycfi { namespace elements
 {
    basic_menu
@@ -147,7 +160,7 @@ namespace cycfi { namespace elements
 
             default:
 #if defined(__APPLE__)
-               mod_ += u8"⇧";
+               mod_ += u8_compat("⇧");
 #else
                mod_ += "Shift+";
 #endif
@@ -157,11 +170,11 @@ namespace cycfi { namespace elements
 
 #if defined(__APPLE__)
       if (mod & mod_alt)
-         mod_ += u8"⌥";
+         mod_ += u8_compat("⌥");
       if (mod & mod_control)
-         mod_ += u8"⌃";
+         mod_ += u8_compat("⌃");
       if ((mod & mod_command) || (mod & mod_action))
-         mod_ += u8"⌘";
+         mod_ += u8_compat("⌘");
 #else
       if ((mod & mod_control) || (mod & mod_action))
          mod_ = "Ctrl+" + mod_;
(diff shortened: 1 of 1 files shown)


djowel:
My instant inclination is to have c++20 code in a separate branch, without any workarounds. We can probably manage the branch to be in sync with the main branches (master and develop), but make it more forward-looking (e.g. concepts!). I haven't really thought this through. The idea has occurred to me just now. Thoughts?

petiaccja:
To be honest, I don't really think having a separate branch is a good idea. The C++20 branch would practically be a pull-request that you have to rebase all the time for years until you can finally merge it. It's a huge maintenance burden, as the rebases will conflict, the commits become irrelevant and disappear, while other commits become fragmented and have to be reordered and squashed to maintain order.

Generally, language features are backwards compatible, and there is enough time between deprecation and removal to move on. This one with the UTF-8 strings is an exception that needs immediate attention, but there are very few of these.

For the new stuff like concepts, maintaining a separate branch would provide few benefits: most of the users would be using the normal branch regardless due to the lack of widespread support for the language feature. When support becomes widespread enough, you can merge the C++20 branch into the main, but that means 2-3 years of pushing dead weight. It's probably better to just delay using concepts altogether by 2-3 years and focusing on other parts of the library.

djowel:
Yes, that's the big downside to it. A big burden. The other possibility is, well, #ifdefs on compiler versions. I guess that's what we can do here. 

Note: The use of #define and reinterpret_cast is another matter:  undesirable and most probably can be reformulated another way. Sorry, but I can't accept a 'temporary fix'.

petiaccja:
> Note: The use of #define and reinterpret_cast is another matter: undesirable and most probably can be reformulated another way. 

Indeed :) To be honest, all the remedies in the paper kinda give me shivers. 

If we think forward, when the whole code base of elements is using C++20, you probably want ```std::u8string```s everywhere, because that's how char8_t was intended. In light of that, it may be a reasonable solution to emulate ```std::u8string``` in C++17:
```c++
#if !defined(_cpp_char8_t)
namespace std {
  using char8_t = char;
  using u8string = std::string;
}
```
Then all code in C++17 can be made to use this alias, and in C++20 mode the alias is simply disabled. In the future, when C++17 is no longer supported, the alias is simply removed from the codebase, and everything keeps working.

The downside is that in C++17, you don't get strong type checking, and you can easily break the C++20 build, but CI should take care of that.

This solution is closely tied to the use of UTF-8 for internal strings, so switching to UTF-32 for all internals would solve the problem equally.

djowel:
> If we think forward, when the whole code base of elements is using C++20, you probably want `std::u8string`s everywhere, because that's how char8_t was intended. In light of that, it may be a reasonable solution to emulate `std::u8string` in C++17:

I think this is a reasonable workaround, except that we should not inject anything into namespace std. It's better to do it the way it's done with filesystem (and other classes somehow not available) in elements, for example:

https://github.com/cycfi/infra/blob/master/include/infra/filesystem.hpp#L37-L46

petiaccja:
It would make it more seamless to transition if it's already in the std namespace, but you are probably right, if somebody has the same "idea" of patching his u8string in then the two libraries will clash.

What do you think about switching to u32string en masse? That would solve this problem as well as make code simpler in general. Of course, overloads for std::string and later std::u8string could be provided next to std::u32string, but those would be immediately converted to u32strings.

djowel:
> What do you think about switching to u32string en masse?

That is what I probably want to do. But I want to sync the work with the "Artist" project to minimize disruptions. This will involve quite a bit of work. Text layout is not easy, to say the least!

Issue #184, opened 2020-07-15
Alternative control method for dial


jpcima:
Hi, there is a user request for a different way of manipulating dials by mouse.

The value is increased or decreased according to the distance from the click origin. (the Manhattan distance I believe)
Moving up and right increases, left and down decreases.

Additionally the dial may have other properties
- it can be associated with a default value, that a click with the middle button can reset to
- a modifier key can allow finer control of the edited value when pressed (shift?)

As I understand, some users of musical software feel more familiar with this paradigm.

djowel:
I knew someone would ask soon :-) Would you like to do some coding? Feel free to add the "manhattan" behavior. I suggest that the rotary or "Manhattan" behavior be a theme variable, so it can be use configurable.

Your other suggestions are also cool! Would you be interested to code that in?

jpcima:
Yes no problem, I try to implement this functionality later today.
This must not be very difficult from what I remember of it.

PR #185, opened 2020-07-15
Add support for linear dials

jpcima:
#184

This work adds support of linear dials. (it's what vstgui calls it)

- tracker info was edited to contain modifier keys, making this information accessible in the event methods
- the theme allows to indicate the mode of dial interaction
- the linear dial range, also in the theme, indicates the pixel distance equivalent to 100% the value

note: I'm not exactly fond of the naming `dial_mode dial_mode;`. I didn't have a better idea.


Xeverous:
> I'm not exactly fond of the naming `dial_mode dial_mode;`

`dial_mode mode`. Whatever better, but do not introduce name conflicts. C++ rules for such shadowing are really complex and I have seen compilers not agreeing on them.

lib/src/element/dial.cpp
```diff
@@ -66,7 +66,27 @@ namespace cycfi { namespace elements
    {
       if (track_info.current != track_info.previous)
       {
-         double new_value = value_from_point(ctx, track_info.current);
+         auto const& theme = get_theme();
+         double new_value;
+
+         if (theme.dial_mode == dial_mode::radial)
+         {
+            new_value = value_from_point(ctx, track_info.current);
```
djowel, line 74: For consistency, I suggest renaming this `radial_value_from_point` and putting the code in the else branch in a `linear_value_from_point function`.

lib/src/support/theme.cpp
```diff
@@ -73,6 +73,9 @@ namespace cycfi { namespace elements
 
     , dialog_button_size         { 100 }
     , message_textbox_size       { { 300, 120 } }
+
+    , dial_mode                  { }
```
djowel, line 77: I suggest making this explicit, instead of default construct. I also suggest making `linear` the default as it seems more common.

djowel:
> note: I'm not exactly fond of the naming `dial_mode dial_mode;`. I didn't have a better idea.

I suggest just using `dial_mode_enum` and we should probably make that naming convention consistent all throughout. Such enums happens all the time!

djowel:
> > note: I'm not exactly fond of the naming `dial_mode dial_mode;`. I didn't have a better idea.
> 
> I suggest just using `dial_mode_enum` and we should probably make that naming convention consistent all throughout. Such enums happens all the time!

Here's an example:

https://github.com/cycfi/elements/blob/master/lib/include/elements/support/canvas.hpp#L110

Here's a non-consistent example that should be renamed:

https://github.com/cycfi/elements/blob/master/lib/include/elements/support/canvas.hpp#L133

lib/src/element/dial.cpp
jpcima: This implementation does not have a direct relation point->value in the linear case.
It's the accumulation of the offsets between one point and the next, with a multiplier applied.
When this takes into account the click origin and the shift modifier, the value will be varied in a stateful manner.

djowel: Ah, of course. Let's just leave this as-is.

djowel:
Looks good to me!

lib/src/element/dial.cpp
jpcima: note: I also look at the way how vstgui does it.
While this is not the carbon copy of how vstgui processes the knob, they have also `valueFromPoint`, which returns the radial value regardless of the knob's setting.

A thing I believe they do, it's if the knob is dragged off-range, you have to turn it back by the same amount to get it back to normal range.
It's not the case in this knob, I have skipped implementing it, as it needs an additional state variable. I didn't consider the difference to be significant enough to bother.

PR #186, opened 2020-07-16
Use HTTPS for the website

redtide:
I noticed that the website works with HTTPS but getting a warning about mixed content because the logo link using HTTP instead.

I've also adapted the `site.url` in `_config.yml`, though I think it should be removed and let Jekyll auto set it based on the current domain in use, e.g. if you'll use `elements.cycfi.com` subdomain and set it in GitHub in a [CNAME file], it will work using it.

As a side note I think it would be better to move the website pages as subdirectory in `cycfi.github.io` repository, at least IMO it doesn't mess up with the code tree, having all projects documentation in the same place.
This is what we end to do with our projects, when in the beginning I thought it was good to have each project with its own documentation in the same place, but realizing that was difficult to manage the resulting website.

[CNAME file]: https://docs.github.com/en/github/working-with-github-pages/managing-a-custom-domain-for-your-github-pages-site#configuring-a-subdomain

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -27,7 +27,7 @@ description: >- # this means to ignore newlines until "baseurl:"
   Elements is a lightweight, fine-grained, HDPI capable, resolution
   independent, modular C++ GUI library.
 baseurl: "/elements" # the subpath of your site, e.g. /blog
-url: "http://cycfi.github.io" # the base hostname & protocol for your site, e.g. http://example.com
+url: "https://cycfi.github.io" # the base hostname & protocol for your site, e.g. http://example.com
 # twitter_username: jekyllrb
 # github_username:  jekyll
 name: "Joel de Guzman"
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -45,7 +45,7 @@
     <main id="content" class="main-content" role="main">
 
         <a href="https://github.com/cycfi/elements">
-            <img src="{{ site.url }}/elements/assets/images/elements.png" alt="Logo" />
+            <img src="assets/images/elements.png" alt="Logo" />
         </a>
 
       {{ content }}


djowel:
> I noticed that the website works with HTTPS but getting a warning about mixed content because the logo link using HTTP instead.

Oh, I didn't notice that! Thanks!
 
> I've also adapted the `site.url` in `_config.yml`, though I think it should be removed and let Jekyll auto set it based on the current domain in use, e.g. if you'll use `elements.cycfi.com` subdomain and set it in GitHub in a [CNAME file](https://docs.github.com/en/github/working-with-github-pages/managing-a-custom-domain-for-your-github-pages-site#configuring-a-subdomain), it will work using it.
> 
> As a side note I think it would be better to move the website pages as subdirectory in `cycfi.github.io` repository, at least IMO it doesn't mess up with the code tree, having all projects documentation in the same place.
> This is what we end to do with our projects, when in the beginning I thought it was good to have each project with its own documentation in the same place, but realizing that was difficult to manage the resulting website.

Go for it. Separate PR perhaps? I'll merge this one for now. Many thanks!

redtide:
I would do if you mean to move the docs directory on the other repo, the only problem is about the git history, do you keep it in elements and restart on the other repository from scratch?

djowel:
> I would do if you mean to move the docs directory on the other repo, the only problem is about the git history, do you keep it in elements and restart on the other repository from scratch?
> 
> Edit: Sorry, I'm afraid I was wrong, if we move there will be difficult to set the CNAME for elements only, but for the entire site if you someday would like to have only elements as subdomain.

I probably misread your initial comments and did not think this through.

redtide:
I was confused and I didn't take into account that you have other pages in your domain, unlike our website that is entirely on GitHub.
So in brief, if you don't plan to use `cycfi.com` (or any other that has already other content) but `github.io` as is now or a dedicated one to point it to, e.g. `cycfi.org` then it would work, at least if keeping the old git history in the current repo and restart in a new one is not a problem.

Xeverous:
> I would do if you mean to move the docs directory on the other repo, the only problem is about the git history, do you keep it in elements and restart on the other repository from scratch?

I'm not sure what @djowel was doing, but (at least in my case) I usually go for one of these 2 strategies:

1. Commit each feature together with its documentation (very consistent docs)

2. Commit docs and code separately (useful when performing gradual docs updates over time)

With 2, all of the commits could easily be moved to a separate repository. 1 is tricky but AFAIK there are some git tricks to only cherry-pick changes in certain paths, `git patch`/`git am` has even an option to subsitute and/or filter paths in any change.

PR #188, opened 2020-07-19
Docs: added anchor headings

redtide:
This [anchor headings] feature is something I use in other Jekyll based website I work on, so I thought it might be useful also for elements website, [here a preview].
It's not like other versions that make appear the anchor on mouse over, it's made with liquid code, without any js code or need for a custom plugin.

This PR is made against develop like the previous one, or it should be better for the documentation to create next PRs against master to make changes already effective?

OT: I would like also to ask what do you think about a PR to add a navigation bar, so it could be easier to choose pages.
In that case I use a HTML file in the `_includes` directory and a YAML file where to add the menu items to make it simple to handle.

[anchor headings]: https://github.com/allejo/jekyll-anchor-headings
[here a preview]: https://redtide.github.io/elements/


djowel:
I like the general idea, but I'm not sure I like the look of having the # there which kinda distracts from the heading. How will the navigation bar look like?

redtide:
Heh, I imagined that the # could be a bit a problem, it can be changed with another character but still visible at any time.

I had in mind my typical flat navigation like at https://sfzformat.com but I forgot a little detail: we don't use bootstrap :-P
So now I'm looking a solution for the cayman theme, will let you know and show a preview.

redtide:
Something like [this](https://redtide.github.io/elements/)? Also a different character for the anchors as test.

djowel:
> Something like [this](https://redtide.github.io/elements/)? Also a different character for the anchors as test.

I think I'd prefer the one that show the anchor on mouse hover. I don't mind js to be honest. Any extra character that always sits there distracts from the main flow of the text.

djowel:
> I had in mind my typical flat navigation like at https://sfzformat.com but I forgot a little detail: we don't use bootstrap :-P
> So now I'm looking a solution for the cayman theme, will let you know and show a preview.

Yeah, that would be nice to have. I'd like to tweak the look and feel once there's something working. Apologies in advance for being finicky with aesthetics :-P But something clean like this: https://bit.ly/3fJV1iL would be great.

redtide:
NP, I'm also like that, maybe I had to make a draft PR as testing more solutions to choose from.
What about the current version? I've made the navbar following the style of the page and centered, and added [anchorjs](https://github.com/bryanbraun/anchorjs).
I think this is the best I can do with my html/css knowledge limits, the alternative would be add bootstrap, which seems to me a bit a overkill for now.

djowel:
> NP, I'm also like that, maybe I had to make a draft PR as testing more solutions to choose from.
> What about the current version? I've made the navbar following the style of the page and centered, and added [anchorjs](https://github.com/bryanbraun/anchorjs).
> I think this is the best I can do with my html/css knowledge limits, the alternative would be add bootstrap, which seems to me a bit a overkill for now.

I love it!

redtide:
ouch I haven't made the changes in the PR, it was only a preview to see if changes was OK, I'll made a fix.

djowel:
> ouch I haven't made the changes in the PR, it was only a preview to see if changes was OK, I'll made a fix.

oops. sorry.

PR #227, opened 2020-09-04
Add Raw Pixel Buffer Support for pixmap

brent-soles:
Adding an interface for taking in a raw buffer of pixels
for a specific format (i.e. unsigned char) which will then render
onto the paint surface.

In addition an example has been added in order to test out the API,
and give a reference as to how to use it.


brent-soles:
Due to personal need, I had to make this addition and thought it would be a good feature to add to the library! Have already used it in a personal project to render a images being buffered to a frame grabber.

I have tested this on Windows/Linux, don't have a access to macOS.

Let me know if there is anything else needed or any thoughts in the PR!

djowel:
Nice! A couple of things:
1. You'd need to adjust this to sync with the develop branch. I believe some changes were made to the contents CMakeLists.txt is written.
2. There's a move to switch elements::canvas to the Artist 2D library, which targets different graphics backends and cairo is no longer a target, which makes things like `CAIRO_FORMAT_ARGB32` no longer possible, and should be abstracted away.
3. Of course, the other targets (Skia and Quartz-2D) should have a way to allow creating raw images from raw pixels, but the interface will have to be cleanly abstracted with no platform/target specific leaking.

We can flesh out the details in the [discord channel](https://discord.gg/NJkadSx).

brent-soles:
Sounds good! Thanks for getting back so quickly.

1. Done!
2. How would you suggest going about targeting the different backends? I haven't looked at the Artist 2D code, is there a good starting point to getting familiar with the API for drawing?
3. Good to know!

brent-soles:
@djowel How do you want to approach this going forward, given the PR for artist?

djowel:
Still thinking about it. I'm almost done with the Artist/Elements port. I'm leaning towards refocusing on that instead. Thoughts?

djowel:
Oh while we're talking, would you have time to do the Quartz2D/MacOS port? It does not have to be perfect, just enough to have it build on MacOS so as not to disrupt the build. I currently have my hands full ATM.

djowel:
I'll close this PR now since it's been applied to artist.

Issue #245, opened 2021-01-06
Example hello_universe, throws exception because image is missing


Andersama:
Seems as though the path given:
```c
   view_.content(
      scroller(image{ "space.jpg" })
   );
```
Is searching inside of the empty example directory, which doesn't have a space.jpg image.

djowel:
Which OS?

Andersama:
Windows 10, on master branch. Seems the issue would be with cmake since that appears to be how the resource directories seem to be set.

djowel:
I just tried. I see no problem with it. When building with CMake, you should get something like this:

<img width="466" alt="Screen Shot 2021-01-07 at 1 46 02 PM" src="https://user-images.githubusercontent.com/568094/103856000-d91e1d00-50ee-11eb-97ca-5d9a9769efe3.png">

The image (space.jpg) is in the 'resources' directory, along with the application fonts.

djowel:
Here, you can see it running: 
![Uploading Screen Shot 2021-01-07 at 1.52.34 PM.png…]()

Andersama:
Here's what I'm getting:
![image](https://user-images.githubusercontent.com/25020235/103876807-f91af580-5088-11eb-8eab-053e9e8902b7.png)

Where the error's clearly stemming from this:
![image](https://user-images.githubusercontent.com/25020235/103876977-44350880-5089-11eb-88e3-854bf2ab4f54.png)

`get_resource_paths()` is returning a directory over. I've not messed with cmake to make sure it's still producing the error.

djowel:
> To clarify: I'm building / running directly in MSVC. If I dig to the output folder and run the .exe that one runs fine*

Yes. It's not an issue with elements. It's a matter of correctly running the app with the correct working directory. It's just the way it is. Think of the exe and all the fonts and resources with it as a kind of a bundle. The app's resources, by default, should be in the same directory as the exe for it to work properly. If that is not desirable, you can add a resource path yourself and tell elements where it is at startup time. See resource_paths.hpp:

https://github.com/cycfi/elements/blob/master/lib/include/elements/support/resource_paths.hpp

djowel:
Feel free to ask more questions! And feel free to close this if you have no further questions.

Andersama:
Ah, well bummer, didn't want to fight cmake in the IDE, definitely will consider using that function at runtime instead.
Absolutely will do.

djowel:
[...]

PR #246, opened 2021-01-08
patch: make sure windows adds resource paths

Andersama:
Should solve: #245


djowel:
Merged. Thanks! But please take note that I changed the target to develop branch, where it will sit for a while before going to master.

Issue #84, opened 2020-04-06
Windows size on Linux (Wayland)


dmacvicar:
I assume this is not normal:

![Screenshot from 2020-04-07 00-53-52](https://user-images.githubusercontent.com/15332/78612850-9df27300-786a-11ea-8461-a1ce3d981208.png)

The window can't be resized.

Opened for tracking. I will look into it  in the next days if time permits.

*Update* Only happens with Wayland backend. `GDK_BACKEND=x11 ./BasicSlidersAndKnobs` works as expected.

djowel:
Oh hrmm... Not normal at all ;-/ But I did check Linux yesterday. I'll check again just to be sure.

djowel:
I tried Ubuntu. I'm getting correct results:

![Screenshot from 2020-04-07 08-20-39](https://user-images.githubusercontent.com/568094/78617211-4aebe080-78a9-11ea-9144-c2bdb73e18b6.png)

dmacvicar:
That is the screenshot I get with

```
GDK_BACKEND=x11 ./BasicSlidersAndKnobs
```

So I would guess this is happening with the Wayland backend.

djowel:
If you can give me exact steps on how to repro this on Ubuntu, I'll see what I can do. I'm a MacOS guy and I know just enough of Linux as needed and I have not gone into the Wayland land yet.

djowel:
OK, I also noticed that I can't resize windows on stock Ubuntu. It seems I have some debugging to do...

djowel:
Alright, I pushed a hotfix to develop. I'll need to merge this hotfix to master as well. Please confirm.

djowel:
> Alright, I pushed a hotfix to develop. I'll need to merge this hotfix to master as well. Please confirm.

Cherry-picked hotfix to master.

dmacvicar:
![Screenshot from 2020-04-07 00-53-52](https://user-images.githubusercontent.com/15332/78813624-97c8d780-79cd-11ea-9dd9-e6980089456e.png)

It is better, but still there is a part covered when the window appears. With lastest develop HEAD I can at least resize, which will uncover the hidden parts, until it starts to recalculate the llayout.

djowel:
> It is better, but still there is a part covered when the window appears. With lastest develop HEAD I can at least resize, which will uncover the hidden parts, until it starts to recalculate the llayout.

This might be a HiDPI issue. Tell me the steps how to repro this.

[...]

PR #250, opened 2021-02-10
A few more GTK fixes

refi64:
This has a more complete solution for #84 + a fix for the redraw glitching on Wayland I had observed.


djowel:
Splendid!!! I suppose this should also be back-ported to the master branch (cairo)?

Issue #254, opened 2021-02-26
Weird caret drawing


Delta-dev-99:
Looking closely to the caret graphics on the text_edit example I noticed that there are actually 2 carets being drawn with different colors and with different visibility intervals. Some screenshots follow.
Both carets shown:
![caret shown](https://user-images.githubusercontent.com/46575220/109252045-95749500-77ba-11eb-8fda-9ade3abd5089.png)
Only one caret shown:
![caret half shown](https://user-images.githubusercontent.com/46575220/109252113-bfc65280-77ba-11eb-9648-3de729161a1b.png)
Both hidden: 
![caret hidden](https://user-images.githubusercontent.com/46575220/109252142-d1a7f580-77ba-11eb-9a53-be9254b8fba9.png)

I failed to capture only the brighter caret on display because it is too bright to see whether or not the other one is drawn, but probably that case happens too.
I actually saw the brighter caret go off, leaving only the dimmer one and after a short interval that one went off too. This proves that there are actually 2 different intervals (or maybe the same but at different phase).
To reproduce, just click around a bit and look carefully to the caret.

--------------------------------------------------------------------------------------------

I also noted some other (short lived) crazy caret behavior after changing caret position. Some screenshots are pretty much self explanatory.
Different caret sizes: ![Untitled](https://user-images.githubusercontent.com/46575220/109253763-78da5c00-77be-11eb-9be1-0c186a9d1660.png) ![Untitled1](https://user-images.githubusercontent.com/46575220/109253766-7aa41f80-77be-11eb-8fce-6a3bda63167f.png)
This one actually shows only the brighter caret (the previously missing case) ![Untitled2](https://user-images.githubusercontent.com/46575220/109253767-7b3cb600-77be-11eb-97e8-d830a470d779.png)
Caret size mismatch (between the two distinct carets simultaneously displayed) ![Untitled3](https://user-images.githubusercontent.com/46575220/109253768-7c6de300-77be-11eb-843e-59ba98e609f2.png)
Some more examples:
![Untitled4](https://user-images.githubusercontent.com/46575220/109253769-7d067980-77be-11eb-8799-939b39a089dc.png)
![Untitled5](https://user-images.githubusercontent.com/46575220/109253772-7d9f1000-77be-11eb-82c4-ed39d71a23bd.png)
Fragmented carets:
![Untitled6](https://user-images.githubusercontent.com/46575220/109253774-7e37a680-77be-11eb-9f66-bd7fdd337b38.png)
[...]

PR #255, opened 2021-02-26
Added an extra 0.5 to caret's horizontal bounds
Partially fixes Weird caret drawing #254

Delta-dev-99:


lib/src/element/text.cpp
```diff
@@ -542,13 +542,16 @@ namespace cycfi { namespace elements
...
+      if (!_is_focus) //No caret if not focused
+         return;
+
       auto& canvas = ctx.canvas;
       auto const& theme = get_theme();
       rect caret_bounds;
       bool has_caret = false;
 
       // Handle the case where text is empty
-      if (_is_focus && _text.empty())
```
djowel, line 551: Why is _is_focus removed? The caret should not be drawn when the text element is not the focus.

djowel: Ignore this. I failed to see the added if statement above.

Delta-dev-99: Ah, sorry about that, I forgot to mention that minor rearrangement.

PR #260, opened 2021-04-03
patch: improve distance still

Andersama:
Unclear as to the exact reason this seems faster without a disassembly of sqrt or sqrtf.
It appears however to be removing two sse cast instructions. This is roughly upwards of 30x faster than the original, although it appears to dip below the previous commit on occasion?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ very much welcome collaboration.
 
 ## News
 
-- 19 January 2020: Added Child Windows to develop (will merge to master in due time). [See Example](https://bit.ly/3iLMJt2).
+- 19 January 2021: Added Child Windows to develop (will merge to master in due time). [See Example](https://bit.ly/3iLMJt2).
 - 25 September 2020: Added [thumbwheels](https://github.com/cycfi/elements/issues/231).
 - 12 September 2020: Linux [Artist 2D Canvas Library](https://github.com/cycfi/artist)
   port feature complete.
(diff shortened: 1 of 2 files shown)


djowel:
It does not really matter. This is an example. For examples, clarity is more important than optimization. The assumption is that the user is expected to do the optimization in non-demo code. For that matter, I'm tempted to revert the previous PR.

Andersama:
Yeah I didn't imagine it would, it's just confusing to me what the assembly was giving back. Since I poked around a bit more in assembly it almost looks as though there's a check for some condition in the sqrt function, I don't know the standard well I'm guessing there's an issue with negative numbers or zero in floating point format because there is a branch instruction buried in both sqrt and sqrtf. I'm sure someone knows the absolute optimal way of going about this, a distance function is a pretty good thing to have for a gui api though. Wish the compiler was smart enough to optimize the pow() call into a multiply on its own, it's the obvious way to write it.

djowel:
> it almost looks as though there's a check for some condition in the sqrt function

It might be some form of optimization. Such things are implementation defined. There are faster versions of pow and sqrt when one does not need total accuracy. In such cases like this, you just want a good approximation and so such alternatives are practical to use. For example, in my DSP library, I use [https://github.com/pmineiro/fastapprox](https://github.com/pmineiro/fastapprox)

Andersama:
Was kind of surprised the bit manipulation trick of the old doom approximation didn't win out, but that's also nowhere near accurate, probably would create issues at small scale.

djowel:
> Was kind of surprised the bit manipulation trick of the old doom approximation didn't win out, but that's also nowhere near accurate, probably would create issues at small scale.

Interesting. For UI handling, issues below 1 pixel do not really matter.

PR #263, opened 2021-04-20
patch: add move_to_back()

Andersama:
A utility function like move_to_front

### DIFF
--- a/lib/include/elements/element/composite.hpp
+++ b/lib/include/elements/element/composite.hpp
@@ -185,6 +185,34 @@ namespace cycfi { namespace elements
                break;
       }
    }
+
+   template<typename C>
+   void move_to_front(C& composite, element_ptr e)
+   {
+      if (e && composite.back() != e)
+      {
+         auto i = std::find(composite.begin(), composite.end(), e);
+         if (i != composite.end())
+         {
+            std::rotate(i, i+1, composite.end());
+            composite.reset();
+         }
+      }
+   }
+
+   template<typename C>
+   void move_to_back(C& composite, element_ptr e)
+   {
+      if (e && composite.front() != e)
+      {
+         auto i = std::find(composite.begin(), composite.end(), e);
+         if (i != composite.end())
+         {
+            std::rotate(composite.begin(), i, i+1);
+            composite.reset();
+         }
+      }
+   }
 }}
 
 #endif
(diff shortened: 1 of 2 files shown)


djowel:
Please check the errors above ^

Andersama:
Ah sorry about that, that's what I get for copy pasting too quickly.

Andersama:
On second look, this is probably a good spot to use std::rotate, don't need to erase and then insert.

djowel:
> On second look, this is probably a good spot to use std::rotate, don't need to erase and then insert.

Makes sense. Do as you please.

djowel:
Maybe you should do the same for move_to_front?

Andersama:
Ah, well good point.

djowel:
While you are at it, how about utilities like move_front and move_back?

djowel:
Also, how about making them generic utility for all composite elements? (not just for top-level view elements). It will probably have this signature:

```
   tempate <typename C>
   void move_to_front(C& composite, element_ptr e);
```

Such utilities can probably be added as free template functions in composite.hpp

Andersama:
Bit lost trying to make sense of composites, it sounds like you know how to get move_to_back and move_to_front working as a utility function.

djowel:
> Bit lost trying to make sense of composites

Composites are just elements in a std::container (currently std::array for fixed sized composites and std::vector for dynamically sized elements). The view `_content` is just one form of composite using `layer` composites: https://cycfi.github.io/elements/layout#layers

`move_to_front` will probably look something like:

```
   tempate <typename C>
   void move_to_front(C& composite, element_ptr e);
   {
      if (e && composite.back() != e)
      {
            auto i = std::find(composite.begin(), composite.end(), e);
            if (i != composite.end())
            {
               std::rotate(i, i+1, composite.end());
               composite.reset();
            }
         }
      }
   }
```

Andersama:
It doesn't look like back(), begin() or end() are members of composite, you want to add those in?

djowel:
> It doesn't look like back(), begin() or end() are members of composite, you want to add those in?

Actual composites are just specializations of std containers. Have a look at:

https://github.com/cycfi/elements/blob/master/lib/include/elements/element/composite.hpp#L116-L119

Note that `array_composite` is `composite<std::array<element_ptr, N>, Base>` which becomes: `class composite : public Base, public std::array<element_ptr, N>`, hence `array_composite` IS-A `std::array<element_ptr, N>`

So, essentially, you deal with composites the same way you deal with std containers.

Andersama:
Well I'm still fighting constructor calls so I didn't get to test if those utility functions do what they should.

djowel:
NP, I'll clean it up later...

PR #288, opened 2022-01-03
On scroll callback and set_position for scrollbars

johannphilippe:
* `std::function<void(point p)> on_scroll` lambda added to scroller_base (in `scroll`, and in `reposition`)
* `set_position(point p)` method added to `scroller_base` allowing to reposition the inside scroller (without calling `on_scroll` callback, since I use it to synchronize several scrollers, it would be infinite call).

### DIFF
--- a/examples/sync_scrollbars/CMakeLists.txt
+++ b/examples/sync_scrollbars/CMakeLists.txt
@@ -24,7 +24,7 @@ endif()
 set(ELEMENTS_APP_PROJECT "SyncScrollbars")
 set(ELEMENTS_APP_TITLE "Sync Scrollbars")
 set(ELEMENTS_APP_COPYRIGHT "Copyright (c) 2016-2020 Joel de Guzman")
-set(ELEMENTS_APP_ID "com.cycfi.empty-starter")
+set(ELEMENTS_APP_ID "com.cycfi.sync-scrollbars")
 set(ELEMENTS_APP_VERSION "1.0")
 
 set(ELEMENTS_APP_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/main.cpp)
--- a/examples/sync_scrollbars/CMakeLists.txt
+++ b/examples/sync_scrollbars/CMakeLists.txt
@@ -4,9 +4,7 @@
 #  Distributed under the MIT License (https://opensource.org/licenses/MIT)
 ###############################################################################
 cmake_minimum_required(VERSION 3.9.6...3.15.0)
-project(EmptyStarter LANGUAGES C CXX)
-
-set(ELEMENTS_ROOT "/home/johann/Documents/Elements/elements")
+project(SyncScrollbars LANGUAGES C CXX)
 
 if (NOT ELEMENTS_ROOT)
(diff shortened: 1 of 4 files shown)


examples/sync_scrollbars/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+set (CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH};${ELEMENTS_ROOT}/cmake")
+
+# If we are building outside the project, you need to set ELEMENTS_ROOT:
+if (NOT ELEMENTS_BUILD_EXAMPLES)
+   include(ElementsConfigCommon)
+   set(ELEMENTS_BUILD_EXAMPLES OFF)
+   add_subdirectory(${ELEMENTS_ROOT} elements)
+endif()
+
+set(ELEMENTS_APP_PROJECT "EmptyStarter")
```
djowel, line 26: EmptyStarter? Shouldn't this be a SyncScrollbars?

examples/sync_scrollbars/main.cpp
```diff
@@ -0,0 +1,56 @@
...
+using namespace cycfi::artist;
+
+// Main window background color
+auto constexpr bkd_color = rgba(35, 35, 37, 255);
+auto background = box(bkd_color);
+
+
+int main(int argc, char* argv[])
+{
+   app _app(argc, argv, "Empty Starter", "com.cycfi.empty-starter");
```
djowel, line 18: Probably bad name?

lib/src/element/port.cpp
```diff
@@ -19,7 +21,7 @@ namespace cycfi { namespace elements
    void port_base::draw(context const& ctx)
    {
       auto state = ctx.canvas.new_state();
-      ctx.canvas.rect(ctx.bounds);
+      ctx.canvas.add_rect(ctx.bounds);
```
djowel, line 24: These are artist APIs, right?

djowel:
As expected, the build is breaking. It seems you targeted the artist library?

johannphilippe:
I made a mistake actually. Since I experiment with artist, I just copied the wrong port and example. It should be fixed in last commit

examples/sync_scrollbars/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+if (NOT ELEMENTS_BUILD_EXAMPLES)
+   include(ElementsConfigCommon)
+   set(ELEMENTS_BUILD_EXAMPLES OFF)
+   add_subdirectory(${ELEMENTS_ROOT} elements)
+endif()
+
+set(ELEMENTS_APP_PROJECT "SyncScrollbars")
+set(ELEMENTS_APP_TITLE "Sync Scrollbars")
+set(ELEMENTS_APP_COPYRIGHT "Copyright (c) 2016-2020 Joel de Guzman")
+set(ELEMENTS_APP_ID "com.cycfi.empty-starter")
```
djowel, line 27: Perhaps this should be "com.cycfi.sync-scrollbars" ?

djowel (changes requested):

PR #295, opened 2022-02-15
Homebrew installation path changed for M1 Macs

ordisbold:
Homebrew is installed below '/opt/homebrew' instead of '/usr/local' for M1 based Macs.

See: https://docs.brew.sh/Installation

As a consequence the CMake script doesn't find libcairo.a. With this change, CMake checks if the HOMEBREW_PREFIX environment variable is set and sets lib_dir accordingly, otherwise falls back to previous behavior.


djowel:
Looks good!

PR #300, opened 2022-03-22
Artist port text input out of range

### LINKED ISSUE #298: input_box bug
The input_box is There is a subscript out of bounds problem in https://github.com/cycfi/elements/blob/ae7b87a8cc53258fac7d1d5a7b96da2994116dcf/lib/src/element/text.cpp#L885
When entering the first words, the _select_end is 1, get_text() value size is 1,out of range.

Here are my changes:

`basic_text_box::caret_metrics info;`
            `if (_select_end >= get_text().size()) {`
                &ensp;&ensp; &ensp;&ensp;  `caret_info(ctx, U"");`
           ` } else {`
               &ensp;&ensp;&ensp; &ensp;  ` caret_info(ctx, &get_text()[_select_end]);`
           ` }`

bizehao:
Fixes #298


bizehao:
Pushed to the wrong branch

PR #303, opened 2022-04-12
fix dynamic_list: subscript out of range in cleanup

PragmaTwice:
When I tried to experience the [active_dynamic_list](https://github.com/cycfi/elements/tree/master/examples/active_dynamic_list) example, I found that clicking the "-" button in my environment (Windows 10, C++ 20, MSVC 2019, latest asio version) would cause the program to crash due to subscript out of bounds.

So I did a simple debug and fixed the problem :)


djowel:
Good catch!

PR #308, opened 2022-08-12
Added horizontal grid lines

dilson-pieve:
Currently, there's a function `vgrid_lines` that creates vertical lines spaced as requested. I added a function `hgrid_lines` that does basically the same, but it creates horizontal lines instead.


djowel:
Wonderful! Thanks for the PR!

PR #311, opened 2022-09-02
Implement X11 host UI

thirtythreeforty:
This PR adds a straight Xlib host UI implementation.  It also provides a couple of amenities for developing plugins, ~namely allowing providing a `native_window_handle` to constructors to allow client code to manage window creation~, and `app::tick()` for running a single UI iteration.

It is still a work in progress.  Notable limitations:

- [x] The other ports do not handle the `native_window_handle` argument to the `window` constructor
- [ ] The clipboard is not implemented
- [x] Double clicks are not yet detected
- [ ] Display scaling is not detected; this may require some opportunistic hacks or DBus calls, I have not researched.
- [ ] IME handling for non-English languages is basically untested (pointers on "easy mode" for testing e.g. pinyin input would be greatly appreciated)
- [x] The macOS port doesn't build because apparently Objective C classes (`NSWindow`) cannot be named from pure C++ code - still figuring this out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ include(ElementsConfigCommon)
 
 option(ELEMENTS_BUILD_EXAMPLES "build Elements library examples" ON)
 option(ELEMENTS_ENABLE_LTO "enable link time optimization for Elements targets" OFF)
-set(ELEMENTS_HOST_UI_LIBRARY "" CACHE STRING "gtk, cocoa or win32")
+set(ELEMENTS_HOST_UI_LIBRARY "" CACHE STRING "gtk, x11, cocoa or win32")
 option(ELEMENTS_HOST_ONLY_WIN7 "If host UI library is win32, reduce elements features to support Windows 7" OFF)
 
 add_subdirectory(lib)
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -142,7 +142,6 @@ elseif (ELEMENTS_HOST_UI_LIBRARY STREQUAL "x11")
    set(ELEMENTS_HOST
       host/x11/app.cpp
       host/x11/base_view.cpp
-      host/x11/key.cpp
       host/x11/window.cpp
    )
 elseif (WIN32)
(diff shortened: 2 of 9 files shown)


djowel:
Looks good! It's a good start. Let's see if we can implement what's missing. Any chance you can have this in the `artist_port` and the Artist lib as well?

djowel:
Hmmm... MacOS failing.

thirtythreeforty:
Yeah, the problem is that NSWindow can't be named from C++, and it forms part of the public API of `window.hpp` (the new feature to pass a native window handle).  When that's built in a C++ (not ObjC) compilation unit, those constructor overloads need to be ifdef'd out.

I can look at an Artist port although my primary objective is getting this port so it can be used for plugin development. I don't see any reason the Cairo version can't be adapted to Artist after the core functionality is stable, right? There's only one or two interaction points with the actual drawing surface, the rest is just X11 boilerplate.

djowel:
I changed the base to a new `gtk_port` branch. `develop` cannot be disrupted by anything unstable. Please fix the error, however you intend to. I'll go review the code from the branch, and maybe develop some code as well.

djowel:
> I can look at an Artist port although my primary objective is getting this port so it can be used for plugin development. I don't see any reason the Cairo version can't be adapted to Artist after the core functionality is stable, right? There's only one or two interaction points with the actual drawing surface, the rest is just X11 boilerplate.

Yes. But back-porting is a bit tricky though, but not impossible. But even if Cairo si not backported, Skia supports X11 just fine. It's in the example driver code, IIRC, rather than GTK.

thirtythreeforty:
Ok, so I've removed the troublesome native window handle change from this patch - it's unrelated to the main X11 bringup - so that should fix the macOS build. I can open a separate PR for it.

djowel:
Hi @thirtythreeforty It's been a while. Do you have any update on this?

thirtythreeforty:
Hey! Let me take a look at the current state of upstream over the next week or two. I am happy to try to push and get it merged.

djowel:
> Hey! Let me take a look at the current state of upstream over the next week or two. I am happy to try to push and get it merged.

Woa! That will be amazing! Take note that the current code base moved now to the long-standing artist_port branch. The old code base is now in the cairo_2023 branch. Both branches are still maintained.

Issue #357, opened 2023-11-17
size_element does not produce consistent bounds


eakoli:
In the 'Size Elements' portion of the documentation it is stated that the 'natural limits of the element will not be violated'.  [Reference](https://cycfi.github.io/elements/layout#semantics-1)

I read this as that given a size (F), and limits (L), the resulting size (S) will be constrained by the limits (L). Such that S is as close to F as possible while still satisfying L.min < S < L.max

Based on this code, size_element adjusts the size to a size that is within the element natural limits. (meeting the above rule)
https://github.com/cycfi/elements/blob/f54953d256e2a218fc8bf1e9926d7cbe552f63ed/lib/include/elements/element/size.hpp#L52-L61

But this code, produces a bounds that violates the elements natural limits.
https://github.com/cycfi/elements/blob/f54953d256e2a218fc8bf1e9926d7cbe552f63ed/lib/include/elements/element/size.hpp#L63-L68

I believe the code should be 

```cpp
template <typename Subject>
inline void size_element<Subject>::prepare_subject(context& ctx)
{
      auto e_limits = this->subject().limits(ctx);
      ctx.bounds.right = ctx.bounds.left + e_limits.min.x;
      ctx.bounds.bottom = ctx.bounds.top + e_limits.min.y;
}
```

The same should apply to vsize_element and hsize_element.

Where this is noticeable is in the following example:

```cpp
int main(int argc, char* argv[])
{
   app _app(argc, argv, "Login", "com.cycfi.text-edit");
   window _win(_app.name());
   _win.on_close = [&_app]() { _app.stop(); };

   view view_(_win);

   view_.content(
            align_center_middle(
                hsize(50,button("a really long title for a button"))
            )
        )
        ,box(colors::white)
   );

   _app.run();
   return 0;
}
```

The resulting button is rendered at its minimum size, but can only be clicked on in the left most 50 pixels.

djowel:
Oh!!! There you go. Good catch. Would you be so kind as to make a PR for this? If it's not too much to ask.

djowel:
> Oh!!! There you go. Good catch. Would you be so kind as to make a PR for this? If it's not too much to ask.

Ah, don't bother. I'll do it. It's just two lines anyway. Many thanks for spotting!

djowel:
Fixed. Again, many thanks! Pushed in `develop` and `cairo_2023` branches.

djowel:
[...]

PR #359, opened 2023-11-20
Fix bounds generation for fixed_size.

eakoli:
Use the fixed w/h as computed by limits() as it apples clamping to ensure the fixed size doest violate the subjects natural limits

This attempts to address the issues noted in https://github.com/cycfi/elements/issues/357

### DIFF
--- a/lib/include/elements/element/size.hpp
+++ b/lib/include/elements/element/size.hpp
@@ -63,7 +63,8 @@ namespace cycfi { namespace elements
    template <typename Subject>
    inline void size_element<Subject>::prepare_subject(context& ctx)
    {
-      auto e_limits = this->subject().limits(ctx);
+      // use the fixed width/height as computed by limits
+      auto e_limits = this->limits(ctx);
       ctx.bounds.right = ctx.bounds.left + e_limits.min.x;
       ctx.bounds.bottom = ctx.bounds.top + e_limits.min.y;
    }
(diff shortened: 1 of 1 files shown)


djowel:
As I mentioned in https://github.com/cycfi/elements/issues/357, I think this is still wrong. Here's a use case:

We have an element with (10 min, 100 max) horizontal limits, and the fixed-size desired is 50. Using only the `min` limits will constrain that to a size of 10 using your code, but 50 is perfectly valid. The correct code is the one I listed in the issue:

```c++
template <typename Subject>
inline void size_element<Subject>::prepare_subject(context& ctx)
{
    auto  e_limits = this->subject().limits(ctx); 
    float size_x = _size.x; 
    float size_y = _size.y; 
    clamp(size_x, e_limits.min.x, e_limits.max.x); 
    clamp(size_y, e_limits.min.y, e_limits.max.y);
    ctx.bounds.right = ctx.bounds.left + size_x; 
    ctx.bounds.bottom = ctx.bounds.top + size_y; 
}
```

djowel:
Also, please change the target to the `cairo_2023` branch.

eakoli:
This code 
```
   auto  e_limits = this->subject().limits(ctx); 
    float size_x = _size.x; 
    float size_y = _size.y; 
    clamp(size_x, e_limits.min.x, e_limits.max.x); 
    clamp(size_y, e_limits.min.y, e_limits.max.y);
```

IS the code that is executed in this->limits(ctx)....thus its logically equivalent, so utilizing this->limits(cxx) produces more DRY code.

djowel:
> IS the code that is executed in this->limits(ctx)....thus its logically equivalent, so utilizing this->limits(cxx) produces more DRY code.

After further investigation, you are correct.

I should really start doing unit testing! It's just that unit testing GUIs is rather tricky.

djowel:
"This branch has conflicts that must be resolved", please target the cairo_2023 branch. Many thanks!

eakoli:
Yea, I was trying to determine a way to do tests for the layout logic. Have some ideas, but nothing feasible yet.

Do you have documented somewhere what the branch pattern is? E.g where things get worked, what represents stable vs not etc.?

PR #360, opened 2023-11-20
Ensure dangling resources are destroyed in move-assignment operator

eakoli:
Fix monitor leaks in move-assignment


djowel:
Please re-target the.`cairo_2023` branch instead. There seems to be some conflicts when I changed it so.

PR #362, opened 2023-11-22
Allow to cancel a timed callback.

na2axl:
As per our discussion in #361, I've implemented a way to cancel a timed callback posted on a view. I've used a simple wrapper that only exposes a `cancel` function.

### DIFF
--- a/lib/include/elements/view.hpp
+++ b/lib/include/elements/view.hpp
@@ -102,8 +102,13 @@ namespace cycfi { namespace elements
       using io_context = asio::io_context;
       io_context&             io();
 
+      struct timer
+      {
+         std::function<void()> cancel;
+      };
+
                               template <typename T, typename F>
-      void                    post(T duration, F f);
+      timer                   post(T duration, F f);
 
                               template <typename F>
       void                    post(F f);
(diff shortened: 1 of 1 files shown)


lib/include/elements/view.hpp
```diff
@@ -102,8 +102,13 @@ namespace cycfi { namespace elements
       using io_context = asio::io_context;
       io_context&             io();
 
+      struct timer
```
djowel, line 105: Perhaps `timer` is not the best name for this? Ah, unless you want to add more timer-related functions later?

na2axl: Yeah that was my feeling. But I start to think even if we add more timer functionalities (maybe for an animation system?) the right place for them won't be there.

I can rename it to sth more contextual, like `timer_canceller`? Or `callback_canceller`?

djowel: How about just returning the timer?

na2axl: I can do that too. I don't know if things can break if the developer use other methods than `cancel`, but I doubt it will. I'll try that

djowel: Asio is stable. It's been in Boost for a VERY long time now. It was actually considered to be in std at one point. It's up to the developer to know and use the asio timer API. But it's quite simple to use.

djowel:
Merged. Thank you, @na2axl !

PR #364, opened 2023-12-01
Implemented a new range_slider

klytje:
I implemented a new range_slider by modifying the existing slider. Most previously single-value functions now returns std::pairs, where the `.first` entry is for the left-most slider, and the `.second`entry is for the right-most slider. The thumbs themselves are returned as std::reference_wrappers. 

- Scrolling does nothing. I don't see how this should intuitively work with two sliders, so I just removed the functionality. I'm not sure what bool scroll() is supposed to return, so I just set it to always return false.  
- You can no longer click anywhere on the track to select a value. You have to specifically click and drag one of the sliders. 
- The two sliders cannot be dragged across eachother, though they _can_  manually be set to be. Trying to move the sliders in this weird configuration results in one being snapped back into place. 
- I made a simple example inspired by my own use for the range_slider. See examples/range_slider. 

I played around a bit with it in the range_slider example, and it seems to work as expected. You probably want to look a bit closer at the code since its my first contribution here though :)


lib/include/elements/element/range_slider.hpp
```diff
@@ -0,0 +1,123 @@
...
+      void                    value_second(double val) { value({_value.first, val});}
+      void                    edit(view& view_, val_type val) override;
+      virtual void            edit_value_first(double val) { value({val, _value.second});}
+      virtual void            edit_value_second(double val) { value({_value.first, val});}
+      void                    edit_value(val_type val) {edit_value_first(val.first); edit_value_second(val.second);}
+
+      rect                    track_bounds(context const& ctx) const;
+      std::pair<rect, rect>   thumb_bounds(context const& ctx) const;
+
+      virtual std::pair<std::reference_wrapper<element const>, std::reference_wrapper<element const>> thumb() const = 0;
```
djowel, line 52: I'd use using typedefs here to keep the formatting tidy, maybe:

using element_ref_pair = std::pair<std::reference_wrapper<element const>, std::reference_wrapper<element const>>;

element_ref_pair       thumb() const = 0;

lib/src/element/range_slider.cpp
```diff
@@ -0,0 +1,257 @@
+/*=============================================================================
+   Copyright (c) 2016-2020 Joel de Guzman
+
```
djowel, line 3: At least add your name here. Now I see it is based on my code. Would be nice to share the copyright with you.

djowel:
> * Scrolling does nothing. I don't see how this should intuitively work with two sliders, so I just removed the functionality. I'm not sure what bool scroll() is supposed to return, so I just set it to always return false.

Maybe move the thumb closest to where the mouse is?

djowel:
> Your library is surprisingly simple to work with, so it went smoothly! I justified spending some time at work to finish it since I need it for my project :)

Thank you! I'm glad you find it easy to work with 👍 :-)

It would be great if you can add that to our list here: https://github.com/cycfi/elements/discussions/352

> 
> > Maybe move the thumb closest to where the mouse is?
> 
> Hmm perhaps? I'm not sure how intuitive this response to scrolling is for a slider like this though. I personally wouldn't expect scrolling to have any effect.

For example, I find it common to hover the cursor over a knob and tdo a wo-finger-scroll to adjust up and down. I think it would be the same behavior if you hover the mouse over the thumb and do a two-finger-scroll. The thumb is rather small, so some tolerance might be good.  I just suggested the closest to where the mouse, but some tolerance might be better (e.g. +10 offset from the thumb bounds), but since the thumb is moving, you'll have to track at first call, which thumb to follow. That's why I simply suggested follow whichever is closest to the mouse.

klytje:
> copyrights

Ah. This is my first time adding copyrights to my code, so thank you for being patient with me there :) 

> It makes me smile when I see someone with c++ skills who know things such as std::decay :-). Outside boost, I am missing the boost community.

Hah I wish I could claim responsibility for that `std::decay`, but I just copied it over from your code ;) 
I rarely work with code templated to such a degree, so I haven't really needed it before. 

> For example, I find it common to hover the cursor over a knob and tdo a wo-finger-scroll to adjust up and down. I think it would be the same behavior if you hover the mouse over the thumb and do a two-finger-scroll. The thumb is rather small, so some tolerance might be good. Some tolerance perhaps (e.g. +10 offset from the thumb bounds). That said, since the thumb is moving, you'll have to track at first call, which thumb to follow. That's why I simply suggested follow whichever is closest to the mouse as a simpler way to do it.

I'm not sure I agree, but I also think I may be too biased towards my own usecase. I'll implement it soon-ish. 

> It would be great if you can add that to our list here: https://github.com/cycfi/elements/discussions/352

I'll add it once I have my article on it published, hopefully in the beginning of 2024. My project will be closed-source until then. It's nowhere near as fancy as those other examples though, haha. I was just informed that most of my userbase will consist of old grumpy biologists who have no idea how to operate a terminal, so I decided to make a simple gui for them.

klytje:
How does scroll tracking work? I see that the `scroll` function is called with the parameter `dir` being the "scroll intensity", and `p` the mouse point. Then you define what appears to be a multiplicative factor `sdir = (1, 1)` for each scroll direction. Why is this step necessary? And what exactly does the `track_scroll(ctx, dir, p)` call accomplish? 

image
Also I believe a small improvement would be to correct the `new_value` to:
`double new_value = value() + (_is_horiz ? dir.x * sdir.x  + !dir.x*dir.y*-sdir.y : dir.y * -sdir.y) * 0.005;` such that you can also scroll horizontal bars with a common "y-only" mouse wheel.

djowel:
> How does scroll tracking work? I see that the `scroll` function is called with the parameter `dir` being the "scroll intensity", and `p` the mouse point. Then you define what appears to be a multiplicative factor `sdir = (1, 1)` for each scroll direction. Why is this step necessary? And what exactly does the `track_scroll(ctx, dir, p)` call accomplish?
> 
> image Also I believe a small improvement would be to correct the `new_value` to: `double new_value = value() + (_is_horiz ? dir.x * sdir.x + !dir.x*dir.y*-sdir.y : dir.y * -sdir.y) * 0.005;` such that you can also scroll horizontal bars with a common "y-only" mouse wheel.

It's all based on feel, testing on separate platforms. Feel is subjective, I know. But historically, scrolls were not really designed for moving controls, for obvious reasons given the name, for example. At one point, I even thought that the dynamics should be platform-dependent because different platforms give different scroll behaviors (the constants). The -1 there should probably take into account the scroll direction, esp. for trackpads (e.g. that "natural" trackpad direction on Macs).

djowel:
> And what exactly does the `track_scroll(ctx, dir, p)` call accomplish?

`track_scroll` ia part of the `tracker` API for scroll movements and allows us to hook into the scroll movements by providing an element::on_tracking function. E.g. it allows us to link two or more controls.

djowel:
oops. conflict? I changed the target of this PR to `main`.

klytje:
> It's all based on feel, testing on separate platforms (...)
> 
> [edit: and indeed `point scroll_direction()` is platform-specific).

Ah okay, that makes sense. If scroll_direction is platform-specific anyway, couldn't it be constexpr?

> oops. conflict? I changed the target of this PR to `main`.

Uh I'm not sure how to solve this - I haven't touched `lib/infra` in any way, and only added to the `examples/CMakeLists.txt`. I assume it's because I forked the `master` branch?

klytje:
I implemented scrolling by introducing a new `state` enum. If a slider overlaps with the mouse position (with 50% larger bounds in each direction), it is selected and moved. Continually scrolling without moving the mouse too much continues to scroll the selected slider. Moving the mouse more than 9 units (pixels?) away terminates the scroll action. I think this feels quite natural to use, actually. 

> Also I believe a small improvement would be to correct the new_value to:
`double new_value = value() + (_is_horiz ? dir.x*sdir.x  + !dir.x*dir.y*-sdir.y : dir.y* -sdir.y) * 0.005;` such that you can also scroll horizontal bars with a common "y-only" mouse wheel.

What'd you think of supporting horizontal scrolling like this? I added it to my own implementation, but I didn't touch the original `slider`.

djowel:
> Ah okay, that makes sense. If scroll_direction is platform-specific anyway, couldn't it be constexpr?
> 
> >

No, it is dynamic. In MacOS here's the implementation:

```
   point scroll_direction()
   {
      float dir = [[[NSUserDefaults standardUserDefaults] objectForKey:@"com.apple.swipescrolldirection"] boolValue]? +1.0f : -1.0f;
      return {dir, dir};
   }
```

djowel:
> > oops. conflict? I changed the target of this PR to `main`.
> 
> Uh I'm not sure how to solve this - I haven't touched `lib/infra` in any way, and only added to the `examples/CMakeLists.txt`. I assume it's because I forked the `master` branch?

Yes. You'll need to rebase as @redtide did.

djowel:
> I implemented scrolling by introducing a new `state` enum. If a slider overlaps with the mouse position (with 50% larger bounds in each direction), it is selected and moved. Continually scrolling without moving the mouse too much continues to scroll the selected slider. Moving the mouse more than 9 units (pixels?) away terminates the scroll action. I think this feels quite natural to use, actually.

Sounds good!
> 
> > Also I believe a small improvement would be to correct the new_value to:
> > `double new_value = value() + (_is_horiz ? dir.x*sdir.x  + !dir.x*dir.y*-sdir.y : dir.y* -sdir.y) * 0.005;` such that you can also scroll horizontal bars with a common "y-only" mouse wheel.
> 
> What'd you think of supporting horizontal scrolling like this? I added it to my own implementation, but I didn't touch the original `slider`.

Go for it! Another PR, perghaps.

PR #367, opened 2023-12-06
Implemented logarithmic slider ticks

klytje:
I implemented a new element `slider_marks_log` and showcase it in the `range_slider` example. 

I couldn't figure out how to do this correctly while limiting myself to the original `<size, num_divs, major_divs>` template, so I switched it around (for this element only) to `<size, major_divs, minor_divs>` instead. For consistency I think either the normal `slider_marks` element should also be updated with this, or alternatively remove the `minor_divs` template from this new element and hardcode it to `10`. I don't see any reason to use a different number than this anyway.

### DIFF
--- a/examples/range_slider/main.cpp
+++ b/examples/range_slider/main.cpp
@@ -13,20 +13,6 @@ auto constexpr bg_color = rgba(35, 35, 37, 255);
 auto constexpr bred     = colors::red.level(0.7).opacity(0.4);
 auto background = box(bg_color);
 
-inline auto pretty_printer = [] (float value) {
-    std::stringstream ss;
-    ss << std::setprecision(3) << value;
-    return ss.str();
-};
-
-inline auto axis_transform = [] (float value) {
-    return value*10;
-};
-
-inline auto axis_transform_inv = [] (float value) {
-    return value/10;
-};
-
 auto make_range_slider(view& _view) {
 	static auto track = basic_track<5, false>(colors::black);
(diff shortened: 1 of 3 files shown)


djowel:
Cool! I've been thinking about that this morning!

djowel:
Screenshots please! :-)

klytje:
[Screencast from 12-06-2023 07:47:17 PM.webm](https://github.com/cycfi/elements/assets/16706862/10bb4edf-5825-4855-a825-8890acb9e248)

Now that I look at it, I think the example could use a little bit of polish with some titles and sections maybe

djowel:
Here's a suggestion:

1. Rewrite `slider_marks` with the updated template params, along with `slider_marks_element`.
2. Rename it to `slider_marks_lin` and `slider_marks_lin_element`
3. Write a forwarding function and alias template `slider_marks` and `slider_marks_element` in terms of `slider_marks_lin` and `slider_marks_lin_element`.
4. Mark these as deprecated with a message referring to the new API.
5. Have the examples use the new API.

djowel:
> @redtide is there a way to swash the commits posthoc via github?

Ah seems there is: https://bit.ly/3uU2EPP

djowel:
> > @redtide is there a way to swash the commits posthoc via github?
> 
> Ah seems there is: https://bit.ly/3uU2EPP

Didn't know it was there all along!

image

PR #368, opened 2023-12-06
Fixed some range_slider issues

klytje:
I fixed the `range_slider` issues we discussed on Discord. More specifically:

-  Scrolling now checks for collisions.
- I added an overload to `range_slider` which takes an additional `float offset_from_center`. This is intended to be in the range [-0.5, 0.5] and specifies where on the shape the ´value´ will be. Setting this to `0.5` allows total overlap of the thumbs, `0` allows exactly no overlap, and `-0.5` blocks movement when the spacing between them is one thumb width.

Please check the new `basic_range_slider_offset_base` and how `basic_range_slider` interacts with it; I'm not entirely sure what I've done there is legal.

Name improvements for `offset_from_center` are happily accepted :) 

Note: this should probably only be accepted after my other PR since this one is based on that.

### DIFF
--- a/examples/buttons/main.cpp
+++ b/examples/buttons/main.cpp
@@ -115,8 +115,8 @@ auto make_buttons(view& view_)
                                  .body_color(bgreen)
                            );
 
-   auto slide_btn1      = slide_button();
-   auto slide_btn2      = slide_button();
+   auto slide_btn1      = slide_switch();
+   auto slide_btn2      = slide_switch();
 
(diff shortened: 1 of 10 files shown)


djowel:
Awesome! This is looking great, @klytje! 

BTW, I've been wanting to ask: can you please do a squash commit to minimize the history?

djowel:
Ooops. Conflicts:
image

klytje:
Should be fixed now :)

PR #372, opened 2023-12-11
Consistent horizontal scrolling

klytje:
I added the possibility of using a vertical scroll wheel to move a horizontal slider in my `range_slider` implementation. This brings this same feature to the 'normal' `slider`.


djowel:
Nice!

PR #395, opened 2024-05-02
CMake version & RangeSlider cleanup

klytje:
I bumped the CMake version to 3.15 as agreed in the Discord. 

I also refactored the RangeSlider example to use smart pointers instead of static elements, and added a tip directing people to check out the Model example to see how to do more advanced input validation.

### DIFF
--- a/examples/range_slider/main.cpp
+++ b/examples/range_slider/main.cpp
@@ -62,6 +62,7 @@ auto make_range_slider = [] (view& _view, auto _range_slider, std::string _title
         } catch (std::exception&) {
             *_min_bg = bred;
         }
+        _view.refresh(*_min_bg);
         return true;
     };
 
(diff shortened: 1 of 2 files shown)


djowel:
Cool! Interestingly, while `master` is the, well, master... `master` is actually tracking `skia_2024`. I work on `skia_2024` first then merge it to `master`. Both are maintained in sync.

djowel:
So... alas, I can't merge this because it is tracking `master` but I need to merge to `skia_2024` first. Odd situation, I know. I'll merge this to master then, then cherry-pick only the latest commits.

examples/range_slider/main.cpp
```diff
@@ -34,53 +34,55 @@ auto make_range_slider = [] (view& _view, auto& _range_slider, std::string _titl
     };
 
     // note: ok to take a reference to _range_slider here, since we declared it static
```
djowel, line 36: Oh, this comment is no longer relevant.

PR #397, opened 2024-05-04
Add busy_bar

Flole998:
It can be used to implement a busy-indicator.


djowel:
There should be a better way. One of the primary rules I adhere to is that the user should not pay for features that he does not use. I might sometimes break that rule if the functionality is used often enough, at least 50% of the time. This is not the case here.

I think this can be better served as a derived class or even a totally separate class which provides the desired behavior, without adding any extra member variable.

Flole998:
@djowel I have changed it now to be a derived class.

djowel:
Looks good! Please keep consistency with formatting code with the one you are basing it with, e.g. 3-spaces tabs, and newline before `{`:

```
foo ()
{
}
```

Instead of 

```
foo () {
}
```

djowel:
Oh, without real tests, we need at least an example. I think it is about time to remove the progress bar example from buttons into its own, with its busy_bar sibling.

djowel:
The code formatting still has some, well, inconsistencies, but barring a code style guideline, which I should do, I suppose I'll just have to tweak it myself, post merge.

Flole998:
If you could come up with a clang_format file that'd allow automated checking and formatting so it's exactly the style you prefer.

I'll see what I can do about the example. The busy bar and progress bar will probably need some buttons to control their behavior, so at least that part can be ripped out of the current example.

djowel:
I think there's a better way to do animation without user intervention, similar to how the caret is drawn by the text editor.

Flole998:
Isn't the point to show, that they can be started and stopped on demand? I think an example to control the busy_bar with a "Start" and "stop" button is more helpful from a programmers point of view, as it's easy to see what code is needed to start/stop it.

djowel:
> If you could come up with a clang_format file that'd allow automated checking and formatting so it's exactly the style you prefer.

Clang format can't deal with my formatting style Haha :-) Not a fan, of automated formatting sorry.
 
> I'll see what I can do about the example. The busy bar and progress bar will probably need some buttons to control their behavior, so at least that part can be ripped out of the current example.

OK, just have something working and I'll do some more work on it.

djowel:
> Isn't the point to show, that they can be started and stopped on demand? I think an example to control the busy_bar with a "Start" and "stop" button is more helpful from a programmers point of view, as it's easy to see what code is needed to start/stop it.

That is correct. Maybe I am just missing the intent of the member functions. BTW, in case you missed it, I am starting to add doxygen docs. Perhapsit makes sense to require code contributions to add doxygen markups as well. Please see the new commits. There's also a style to be followed for consistency.

PR #401, opened 2024-05-18
Expansion of the rounded buttons

klytje:
With this PR, we can now specify the 'roundedness' of each individual corner of a button. 

There are now four ways to control them:
1. Using the existing `corner_radius(r)` method to specify all corners at once,
2. Using the new overloaded version `corner_radius(r1, r2, r3, r4)` to specify them indivdually,
3. Using the new `rounded_*(r)` series of methods to specify a half-rounded side,
4. Using the new `rounded_corner_*(r)` series of methods to specify individual corners.

I expanded the `Buttons` example to showcase this new functionality. 

I also added a missing implementation of the `gen_text_color text_color(color color_)` method I discovered during my testing. Previously it could not compile code specifying the text color of buttons.


lib/src/support/canvas.cpp
```diff
@@ -247,6 +247,27 @@ namespace cycfi { namespace elements
...
+      auto r = r_.right;
+      auto b = r_.bottom;
+      auto const a = M_PI/180.0;
+      for (auto& r : radius) {
+         r = std::min(r, std::min(r_.width(), r_.height()) / 2);
+      }
+
+      // radii array follow the unit circle defined counterclockwise with 0 at (x, y) = (1, 0)
+      // we have to map this to the cairo definition
+      cairo_new_sub_path(&_context);
```
djowel, line 263: This is Cairo specific. There should be equivalent canvas functions so it can port to skia directly.

lib/include/elements/element/gallery/button.hpp
```diff
@@ -231,26 +240,97 @@ namespace cycfi { namespace elements
...
+
+      float                  _radius;
+   };
+
+   template <typename Base>
+   struct button_styler_with_individual_corner_radii : Base
+   {
+      using base_type = button_styler_with_individual_corner_radii<typename Base::base_type>;
+
+                              button_styler_with_individual_corner_radii(Base base, float top_left, float top_right, float bottom_right, float bottom_left)
```
djowel, line 288: Please limit to 80 characters :-)

lib/src/support/canvas.cpp
klytje: I just followed the implementation of the existing `add_round_rect` method. Should this be ported to skia or redefined in terms of existing methods? Considering this is a more general version of the existing method, I think it would make sense to just replace the old one with this.

djowel: > I just followed the implementation of the existing `add_round_rect` method. Should this be ported to skia or redefined in terms of existing methods? Considering this is a more general version of the existing method, I think it would make sense to just replace the old one with this.

No. Skia and Quartz-2D already have round_rect. Cairo does not. What you are seeing is a workaround for the lack of native round_rect facility in Cairo. 

Elements itself (apart from canvas, which is ported to artist for different backends), should use only canvas facilities. have a look at the canvas arc functions: https://github.com/cycfi/elements/blob/master/lib/include/elements/support/canvas.hpp#L65-L71

djowel: > I think it would make sense to just replace the old one with this.

Canvas, and hence Artist, follows the W3C canvas API. round rect is a specialized API which assumes a single radius for all corners. If you want something more specialized than that, then use primitive lines and arcs and even splines.

djowel: In general, any new code or code change should also compile, build and run with the `skia_2024` branch. Please make sure of that.

klytje: That makes sense. I didn't know the canvas api was fixed.

lib/include/elements/element/gallery/button.hpp
klytje: While I didn't quite manage to keep it below 80 characters, I've now reduced it to be consistent with the size of the existing code (longest is ~100 chars).

djowel: > While I didn't quite manage to keep it below 80 characters, I've now reduced it to be consistent with the size of the existing code (longest is ~100 chars).

Yes, it's not a strict rule, but more like a guideline. ~100 chars is fine.

lib/src/support/canvas.cpp
djowel: > That makes sense. I didn't know the canvas api was fixed.

Yeah, sorry it was not documented (yet). Working on it.

djowel:
Looking good. Is this ready for merge? I see some commented out code.

lib/src/support/draw_utils.cpp
```diff
@@ -98,16 +117,19 @@ namespace cycfi { namespace elements
       cnv.fill_style(gradient);
 
       cnv.begin_path();
-      cnv.add_round_rect(bounds.inset(1, 1), corner_radius-1);
+      draw_round_rect(cnv, bounds.inset(1, 1), {corner_radius[0]-1, corner_radius[1]-1, corner_radius[2]-1, corner_radius[3]-1});
```
djowel, line 120: I think `bounds.inset(1, 1)` ought to be a temporary: `auto r = bounds.inset(1, 1)`, which will shorten the line as well.

lib/src/support/draw_utils.cpp
```diff
@@ -79,7 +79,26 @@ namespace cycfi { namespace elements
...
+   {
+      auto l = bounds.left;
+      auto t = bounds.top;
+      auto r = bounds.right;
+      auto b = bounds.bottom;
+      for (auto& r : radius) {
+         r = std::min(r, std::min(bounds.width(), bounds.height()) / 2);
+      }
+
+      // radii array follow the unit circle defined counterclockwise with 0 at (x, y) = (1, 0)
```
djowel, line 92: Isn't it more natural for the radii to follow a clockwise path? This follows the natural API of `cnv.arc(...)` as well, which defaults to a clockwise path.

lib/src/support/draw_utils.cpp
```diff
@@ -98,16 +117,19 @@ namespace cycfi { namespace elements
       cnv.fill_style(gradient);
 
       cnv.begin_path();
-      cnv.add_round_rect(bounds.inset(1, 1), corner_radius-1);
+      draw_round_rect(cnv, bounds.inset(1, 1), {corner_radius[0]-1, corner_radius[1]-1, corner_radius[2]-1, corner_radius[3]-1});
+      // cnv.add_round_rect(bounds.inset(1, 1), {corner_radius[0]-1, corner_radius[1]-1, corner_radius[2]-1, corner_radius[3]-1});
```
djowel, line 121: We can remove these commented out code.

lib/src/support/draw_utils.cpp
```diff
@@ -98,16 +117,19 @@ namespace cycfi { namespace elements
...
 
       cnv.fill_style(gradient);
       cnv.fill();
 
       cnv.begin_path();
-      cnv.add_round_rect(bounds.inset(0.5, 0.5), corner_radius-0.5);
+      draw_round_rect(cnv, bounds.inset(0.5, 0.5), {corner_radius[0]-0.5f, corner_radius[1]-0.5f, corner_radius[2]-0.5f, corner_radius[3]-0.5f});
```
djowel, line 131: DItto: `r = bounds.inset(0.5, 0.5)`

lib/src/support/draw_utils.cpp
klytje: Personally I find the anti-clockwise path _much_ more natural since that's how the unit circle is defined. I was actually very surprised to learn this wasn't the standard in the API. 

I don't have a strong preference if you prefer the other way around though.

klytje: Perhaps we should just add a small struct with named fields to avoid future confusion.

djowel: > Perhaps we should just add a small struct with named fields to avoid future confusion.

This, I think, a good solution. However, when passing in a struct the usual way using brace initializer, you will still pass it in sequential order. We still have to choose the right sequence.

klytje: That's true. I think with a constructor illustrating the order it'll be less important though. 

Anyway, my main argument for the current implementation is that it follows the unit circle, which in turn is the conventional way to define a spherical coordinate system. Again I gotta say I am extremely surprised the W3C API does not follow this convention (I think this is perhaps because the y-axis is flipped?).

On the other hand, perhaps the connection to spherical coordinates is too vague for this connection to hold. I am thinking of this problem in terms of the four arcs we have to draw, so it is an obvious connection for me to make.

djowel: > I don't have a strong preference if you prefer the other way around though.

After thinking about it some more, I think I'd go with a clockwise direction if that's OK with you.

klytje:
@djowel I added a small corner_radii struct, with initialization order: `bottom_right`, `bottom_left`, `top_left`, `top_right`. Same as the API I believe.

djowel:
> @djowel I added a small corner_radii struct, with initialization order: `bottom_right`, `bottom_left`, `top_left`, `top_right`. Same as the API I believe.

Hi @klytje I think the correct sequence is top_left, top_right, bottom_right, bottom_left :-)

Interestingly, if you ask ChatGPT to make an API for drawing a round rectangle with customizable corner radii, it's the one it will recommend:

```
void draw_round_rectangle(float x, float y, float width, float height,
                          float radius_tl, float radius_tr,
                          float radius_br, float radius_bl);
```

Then, I asked: "why did you specify the radii from top-left going clockwise?"

ChatGPT follows the large language model from a vast source of programming code, and its answer is:

"Specifying the radii from the top-left corner going clockwise is a common convention in graphics programming. It follows a logical order that aligns with how many people read and interpret shapes and diagrams."

It is indeed customary in computer graphics to give precedence to the top-left, and in general, the horizontal coordinate first (i.e. x, before y).

klytje:
@djowel I implemented it according to the W3C docs:
image
_"The points at startAngle and endAngle this circle's circumference, measured in radians clockwise from the positive x-axis, are the start and end points respectively."_

Indeed, looking at the my implementation you'll see the angle going from 0 --> 90 --> 180 --> 270 --> 360 as it is now:
https://github.com/klytje/elements/blob/94f8101be8cb758328d417808fc98a6a606cafd8/lib/src/support/draw_utils.cpp#L99-L104

Your order is what I meant as the third option, following the `rect`, though looking back it wasn't super clear that was what I meant.

klytje:
Also I agree that starting from the top-left feels more natural for most objects, _except_ circles :)

djowel:
> > @djowel I added a small corner_radii struct, with initialization order: `bottom_right`, `bottom_left`, `top_left`, `top_right`. Same as the API I believe.
> 
> Hi @klytje I think the correct sequence is top_left, top_right, bottom_right, bottom_left :-)
> 
> Interestingly, if you ask ChatGPT to make an API for drawing a round rectangle with customizable corner radii, it's the one it will recommend:
> 
> ```
> void draw_round_rectangle(float x, float y, float width, float height,
>                           float radius_tl, float radius_tr,
>                           float radius_br, float radius_bl);
> ```
> 
> Then, I asked: "why did you specify the radii from top-left going clockwise?"
> 
> ChatGPT follows the large language model from a vast source of programming code, and its answer is:
> 
> "Specifying the radii from the top-left corner going clockwise is a common convention in graphics programming. It follows a logical order that aligns with how many people read and interpret shapes and diagrams."
> 
> It is indeed customary in computer graphics to give precedence to the top-left, and in general, the horizontal coordinate first (i.e. x, before y).

BTW, skia goes so far as specify the horizontal and vertical radii for all corners:

https://api.skia.org/classSkRRect.html

klytje:
Changed initialization order to `top_left`, `top_right`, `bottom_right`, `bottom_left`.

djowel:
> Changed initialization order to `top_left`, `top_right`, `bottom_right`, `bottom_left`.

Thanks for your patience! :-)

lib/include/elements/element/gallery/button.hpp
```diff
@@ -36,6 +36,8 @@ namespace cycfi { namespace elements
 
       enum icon_placement {icon_none, icon_left, icon_right};
       enum label_alignment {align_left, align_center, align_right};
+      enum class direction {up, down, left, right};
+      enum class corner {top_left, top_right, bottom_left, bottom_right};
```
djowel, line 40: I do not use `enum class` for enums in a class. It is redundant and is unnecessarily verbose. The containing class is already a good qualifier.

Also, I think the sequence should also follow the clockwise flow, so: 

```
enum corner {top_left, top_right, bottom_right, bottom_left};
```
Just in case; this can also be used as an index.

djowel:
Ouch. I forgot. This should be against skia_2024 first. Anyway, I'll just do the cherry-pick dance again. No worries.

djowel:
Oh gee... It seems Skia does arcs differently –not connecting the lines between the arcs!

djowel:
> Oh gee... It seems Skia does arcs differently –not connecting the lines between the arcs!

At least Quartz2D is doing it correctly! I'll push to a`skia_2024_dev` branch. See if you can fix the Artist's Skia implementation of arc to follow W3C's behavior.

djowel:
BTW, in the demo cpp, I think we should use grids to align the 3 R/G/B buttons properly.

djowel:
> > Oh gee... It seems Skia does arcs differently –not connecting the lines between the arcs!
> 
> At least Quartz2D is doing it correctly! I'll push to a`skia_2024_dev` branch. See if you can fix the Artist's Skia implementation of arc to follow W3C's behavior.

FYI: I have a fix for skia now.

klytje:
> FYI: I have a fix for skia now.

Ah cool! I gave it a quick glance this morning but realized I'd have to dedicate more time than I had to fix it :)

djowel:
> > FYI: I have a fix for skia now.
> 
> Ah cool! I gave it a quick glance this morning but realized I'd have to dedicate more time than I had to fix it :)

Getting there...

image

LOL :-)

djowel:
OK fixed. I'll merge to skia_2024

djowel:
> OK fixed. I'll merge to skia_2024

Merged

djowel:
@klytje... Reopening, I think rounded_left is wrong:

image

If you zoom in, the right-top and right-bottom corners are not sharp. It is fixed by using `.corner_radius(10, 0, 0, 10)`

Or is it intentional? Assuming the default radius when not specified? It does not seem right, no?

PR #406, opened 2024-05-30
Allow disabled checkboxes to be checked in code.

eisen:
This is for instances when a checkboxes should be checked but the user won't be able to uncheck it.

### DIFF
--- a/examples/buttons/main.cpp
+++ b/examples/buttons/main.cpp
@@ -228,11 +228,14 @@ auto make_controls(view& view_)
    auto  check_box2 = check_box("The Nexus Meridian Unfolding");
    auto  check_box3 = check_box("Serenity Dreamscape Exploration");
    auto  check_box4 = check_box("Forever Disabled");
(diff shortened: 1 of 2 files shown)


djowel:
Could you expound a bit more on this PR, maybe with some screenshots. A use-case would be great to grasp the idea.

eisen:
Sure. I added an example on the buttons example too. I'm building a tool to make it easier to build engines using Vulkan for for computer graphics researchers. During setup, there are extensions that the user can enable. And there are a few that need to be enabled no matter what. These extensions I want to be displayed as checked but disabled so the user can't uncheck them as they are required.

eisen:
image

djowel:
> Sure. I added an example on the buttons example too. I'm building a tool to make it easier to build engines using Vulkan for computer graphics researchers. During setup, there are extensions that the user can enable. And there are a few that need to be enabled no matter what. These extensions I want to be displayed as checked but disabled so the user can't uncheck them as they are required.

Wonderful!

PR #411, opened 2024-06-22
Add coverity scans

Flole998:
This requires a new environment to be created with the token and email address from coverity. Due to that I was not able to test this. I think you could put this into a separate branch and then then use the manual trigger to test it.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,78 @@
+name: Coverity
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


djowel:
> I think you could put this into a separate branch

Why can't we just try it on the skia_2024 branch?

Flole998:
Sure, I just wanted to avoid needing multiple commits on that branch to get a working version. If you don't mind that, using the skia_2024 branch is perfectly fine.

PR #416, opened 2024-06-26
Unified layout and limits functions for vtile and htile.

bruno-j-nicoletti:
Added functions to get values out of rect, point and view_stretch by axis.

Refactored the tile code so that the limit and layout method on vtile_element and htile_element call common templated code using that axis dependent code. Gives single code path for both.

### DIFF
--- a/lib/include/elements/support/point.hpp
+++ b/lib/include/elements/support/point.hpp
@@ -24,6 +24,9 @@ namespace cycfi { namespace elements
       constexpr point   move(float dx, float dy) const;
       constexpr point   move_to(float x, float y) const;
 
+      constexpr float   coord(bool is_x_axis) const;
+      constexpr float&  coord(bool is_x_axis);
+
       float             x;
       float             y;
    };
@@ -78,6 +81,16 @@ namespace cycfi { namespace elements
       r.x = x_;
       r.y = y_;
       return r;
+   }
+
+   inline constexpr float point::coord(bool is_x_axis) const
+   {
+     return is_x_axis ? x : y;
+   }
+
+   constexpr float& point::coord(bool is_x_axis)
+   {
+     return is_x_axis ? x : y;
    }
 }}
 
--- a/lib/include/elements/support/rect.hpp
+++ b/lib/include/elements/support/rect.hpp
@@ -53,6 +53,12 @@ namespace cycfi { namespace elements
       constexpr rect    move_to(float x, float y) const;
       constexpr rect    inset(float x_inset = 1.0, float y_inset = 1.0) const;
 
+      constexpr float   axisDelta(bool is_x_axis) const;
+      constexpr float   axisMin(bool is_x_axis ) const;
+      constexpr float   axisMax(bool is_x_axis ) const;
+      constexpr float&  axisMin(bool is_x_axis );
+      constexpr float&  axisMax(bool is_x_axis );
+
       float             left;
       float             top;
       float             right;
@@ -206,6 +212,32 @@ namespace cycfi { namespace elements
          clear(r);
 
       return r;
+   }
+
+
+   constexpr float rect::axisDelta(bool is_x_axis) const
+   {
+     return is_x_axis ? width() : height();
+   }
+
+   constexpr float rect::axisMin(bool is_x_axis ) const
+   {
+     return is_x_axis ? left : top;
+   }
+
+   constexpr float rect::axisMax(bool is_x_axis ) const
+   {
+     return is_x_axis ? right : bottom;
+   }
+
+   constexpr float& rect::axisMin(bool is_x_axis )
+   {
+     return is_x_axis ? left : top;
+   }
+
+   constexpr float& rect::axisMax(bool is_x_axis )
+   {
+     return is_x_axis ? right : bottom;
    }
 
    constexpr bool is_valid(rect r)
--- a/lib/src/element/tile.cpp
+++ b/lib/src/element/tile.cpp
@@ -90,6 +90,71 @@ namespace cycfi::elements
          std::sort(elements.begin(), elements.end(),
             [](layout_info lhs, layout_info rhs){ return lhs.index < rhs.index; });
       }
+
+     template<bool IS_X_AXIS, class TILE_ELEMENT>
+     [[gnu::always_inline]] inline auto compute_limits(basic_context const& ctx, const TILE_ELEMENT &tile) -> view_limits
+     {
+      view_limits limits{{0.0, 0.0},
+                         {IS_X_AXIS ? 0.0 : full_extent,
+                          IS_X_AXIS ? full_extent : 0.0}};
+      for (std::size_t i = 0; i != tile.size();  ++i)
+      {
+         auto el = tile.at(i).limits(ctx);
+
+         limits.min.coord(IS_X_AXIS) += el.min.coord(IS_X_AXIS);
+         limits.max.coord(IS_X_AXIS) += el.max.coord(IS_X_AXIS);
+         clamp_min(limits.min.coord(! IS_X_AXIS), el.min.coord(! IS_X_AXIS));
+         clamp_max(limits.max.coord(! IS_X_AXIS), el.max.coord(! IS_X_AXIS));
+      }
+
+      clamp_min(limits.max.coord(! IS_X_AXIS), limits.min.coord(! IS_X_AXIS));
+      clamp_max(limits.max.coord(IS_X_AXIS), full_extent);
+      return limits;
+     }
+
+     template<bool IS_X_AXIS, class TILE_ELEMENT>
+     [[gnu::always_inline]] inline auto compute_layout(context const& ctx, TILE_ELEMENT &tile, std::vector<float> &tilePos) -> void
+     {
+       auto const sz = tile.size();
+
+       // Collect min, max, and stretch information from each element.
+       // Initially set the allocation sizes of each element to its minimum.
+       std::vector<layout_info> info(sz);
+       for (std::size_t i = 0; i != sz; ++i)
+       {
+         auto& elem = tile.at(i);
+         auto limits = elem.limits(ctx);
+         info[i].stretch = elem.stretch().coord(IS_X_AXIS);
+         info[i].min = limits.min.coord(IS_X_AXIS);
+         info[i].max = limits.max.coord(IS_X_AXIS);
+         info[i].alloc = limits.min.coord(IS_X_AXIS);
+         info[i].index = i;
+       }
+
+       auto const otherMin = ctx.bounds.axisMin(not IS_X_AXIS);
+       auto const otherMax = ctx.bounds.axisMax(not IS_X_AXIS);
+       auto const myMin = ctx.bounds.axisMin(IS_X_AXIS);
+       auto const myDelta = ctx.bounds.axisDelta(IS_X_AXIS);
+       // Compute the best fit for all elements
+       allocate(myDelta, info);
+
+       // Now we have the final layout. We can now layout the individual
+       // elements.
+       tilePos.resize(sz);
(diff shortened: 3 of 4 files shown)


lib/include/elements/support/point.hpp
```diff
@@ -24,6 +24,9 @@ namespace cycfi { namespace elements
       constexpr point   move(float dx, float dy) const;
       constexpr point   move_to(float x, float y) const;
 
+      constexpr float   coord(bool is_x_axis) const;
```
bruno-j-nicoletti, line 27: Would an operator [] be useful here as well?
`constexpr float operator[](bool is_x_axis) const`

djowel:
Looking at this a bit, I think it can be done non-intrusively of `point` and `rect`, via free functions, if you treat `point` and `rect` as fixed sized arrays of floats via `operator[](std:;size_t)`. The only change will be the addition of `operator[]`

bruno-j-nicoletti:
> Looking at this a bit, I think it can be done non-intrusively of `point` and `rect`, via free functions, if you treat `point` and `rect` as fixed sized arrays of floats via `operator[](std::size_t)`. The only change will be the addition of `operator[]`

That's how I normally write my vector/point classes, not that I do that any more, currently my code uses the Eigen library, https://gitlab.com/libeigen/eigen. The only issue is then is bounds checking, which a bool saves you from in the 2D case. 

To me, logically treating bounding boxes/rects as flat arrays doesn't make as much sense as it does for vectors/points. It's two point values representing the min/max values on the axii in your N dimensional space. 

Happy to flip it to free functions.

bruno-j-nicoletti:
Ha! You edited your comment as I was posting. My other suggestion was for an axis enum to do the indexing into points/rects as opposed to a bool.

djowel:
> Ha! You edited your comment as I was posting. My other unstatate idea was for an axis enum to do the indexing into points/rects as opposed to a bool.

Yes, sorry. my mind was wandering as I explored some more.

djowel:
> Happy to flip it to free functions.

Yes. The general rule is ... If it can be a free function, make it so.

djowel:
> The only issue is then is bounds checking, which a bool saves you from in the 2D case.

True! And that is why I took it back :-)

bruno-j-nicoletti:
Only issue then is the axis specifiying rect constructor. 

`constexpr         rect::rect(axis a, float this_axis_min, float other_axis_min, float this_axis_max, float other_axis_max);`

Given it's creating a rect I'd argue this shouldn't be a free function.

djowel:
> Only issue then is the axis specifiying rect constructor.
> 
> `constexpr rect::rect(axis a, float this_axis_min, float other_axis_min, float this_axis_max, float other_axis_max);`
> 
> Given it's creating a rect I'd argue this shouldn't be a free function.

constexpr  rect make_rect(axis a, float this_axis_min, float other_axis_min, float this_axis_max, float other_axis_max);

bruno-j-nicoletti:
I'm removing my camelHump method names and refactoring it now.

Looking at the code I'd argue for an pair of `point::operator[](axis)` methods, I think the version with the [] operator is clearer code as it's a familiar paradigm for coding.

```
     // Free function
     coord(limits.min, AXIS) += coord(el.min, AXIS);

     // operator
     limits.min[AXIS] += el.min[AXIS];
```
You can't write [] operators as free functions.

djowel:
> Looking at the code I'd argue for an pair of `point::operator[](axis)` methods, I think the version with the [] operator is clearer code as it's a familiar paradigm for coding.

Go for it!

djowel:
> > Looking at the code I'd argue for an pair of `point::operator[](axis)` methods, I think the version with the [] operator is clearer code as it's a familiar paradigm for coding.
> 
> Go for it! I like it!

Except for the ALLCAPS :-P Haha.

djowel:
> I'm removing my camelHump method names and refactoring it now.

WeDontNeedEmHumps :-)

bruno-j-nicoletti:
snakes_are_evil!

bruno-j-nicoletti:
> > > Looking at the code I'd argue for an pair of `point::operator[](axis)` methods, I think the version with the [] operator is clearer code as it's a familiar paradigm for coding.
> > 
> > 
> > Go for it! I like it!
> 
> Except for the ALLCAPS :-P Haha.

It's a template arg, I'll flip it to your stylee.

djowel:
> snakes_are_evil!

HumpsAreForTemplateTypenames :-) We love evil_snakes everywhere.

djowel:
> It's a template arg, I'll flip it to your stylee.

If it's a non-type template arg, then it's also an evil_snake :-)

djowel:
> > snakes_are_evil!
> 
> HumpsAreForTemplateTypenames :-) We love evil_snakes everywhere else.

https://bit.ly/4bo0mID

bruno-j-nicoletti:
> > > snakes_are_evil!
> > 
> > 
> > HumpsAreForTemplateTypenames :-) We love evil_snakes everywhere else.
> 
> https://bit.ly/4bo0mID

If we are being pedantic, from that thread....

> Never start an identifier with an underscore, and don't have a double underscore anywhere inside an identifier. Those are reserved for the implementation.

My member variuables end with an underscore as opposed to starting with them.

:-P

bruno-j-nicoletti:
Done!

djowel:
> My member variuables end with an underscore as opposed to starting with them.
> 
> :-P

That is actually wrong. The standard says never start with TWO underscores. But also never start a underscore and a cap, like _BAD

djowel:
> > My member variuables end with an underscore as opposed to starting with them.
> > :-P
> 
> That is actually wrong. The standard says never start with TWO underscores. But also never start a underscore and a cap, like _BAD

Basically...

The use of two underscores (` __ ') in identifiers is reserved for the compiler's internal use according to the ANSI-C standard. Underscores (` _ ') are often used in names of library functions (such as " _main " and " _exit "). In order to avoid collisions, do not begin an identifier with an underscore.

BUT!!! C++ is scoped. Take that recommendation as old-fashioned, pre namespaces and strict scopes.

djowel:
> > > > My member variuables end with an underscore as opposed to starting with them.
> > > > :-P
> > > 
> > > 
> > > That is actually wrong. The standard says never start with TWO underscores. But also never start a underscore and a cap, like _BAD
> > 
> > 
> > Basically...
> > The use of two underscores (` __ ') in identifiers is reserved for the compiler's internal use according to the ANSI-C standard. Underscores (` _ ') are often used in names of library functions (such as " _main " and " _exit "). In order to avoid collisions, do not begin an identifier with an underscore.
> > BUT!!! C++ is scoped. Take that recommendation as old-fashioned, pre namespaces and strict scopes.
> 
> I learn something new everyday.

Summary:

1. Reserved in any scope, including for use as macros:
 - identifiers beginning with an underscore followed immediately by an uppercase letter
 - identifiers containing adjacent underscores (or "double underscore")
 
2. Reserved in the global namespace:
 - identifiers beginning with an underscore

lib/src/element/tile.cpp
```diff
@@ -90,68 +90,94 @@ namespace cycfi::elements
          std::sort(elements.begin(), elements.end(),
             [](layout_info lhs, layout_info rhs){ return lhs.index < rhs.index; });
       }
+
+     template<axis Axis, class TileElement, axis OtherAxis = other(Axis)>
+     [[gnu::always_inline]] inline auto compute_limits(basic_context const& ctx, const TileElement &tile) -> view_limits
```
djowel, line 95: Use CYCFI_FORCE_INLINE (defined in infra/support.hpp).

bruno-j-nicoletti: done

bruno-j-nicoletti:
Done

lib/include/elements/support/rect.hpp
```diff
@@ -235,6 +235,41 @@ namespace cycfi { namespace elements
       r.right = 0.0;
       r.bottom = 0.0;
    }
+
+   inline constexpr float axis_delta(rect const &r, axis a)
+   {
+     return a == axis::x ? r.width() : r.height();
+   }
```
djowel, line 242: Maybe `axis_extent` is a better name?

lib/include/elements/support/rect.hpp
```diff
@@ -235,6 +235,41 @@ namespace cycfi { namespace elements
...
+
+   inline constexpr float& axis_max(rect &r, axis a)
+   {
+     return a == axis::x ? r.right : r.bottom;
+   }
+
+   inline constexpr auto make_rect(axis a, float this_axis_min, float other_axis_min, float this_axis_max, float other_axis_max) -> rect
+   {
+     if(a == axis::x) {
+       return rect(this_axis_min, other_axis_min, this_axis_max, other_axis_max);
```
djowel, line 267: minor: space after constructs like if, switch, etc., to distinguish them from functions.

lib/src/element/tile.cpp
```diff
@@ -90,68 +90,94 @@ namespace cycfi::elements
          std::sort(elements.begin(), elements.end(),
             [](layout_info lhs, layout_info rhs){ return lhs.index < rhs.index; });
       }
+
+     template<axis Axis, class TileElement, axis OtherAxis = other(Axis)>
```
djowel, line 94: space after template for consistency: `template <...>`

lib/src/element/tile.cpp
```diff
@@ -90,68 +90,94 @@ namespace cycfi::elements
...
+     CYCFI_FORCE_INLINE auto compute_limits(basic_context const& ctx, const TileElement &tile) -> view_limits
```
djowel, line 95: 3-spaces indents

lib/src/element/tile.cpp
```diff
@@ -90,68 +90,94 @@ namespace cycfi::elements
...
+     {
+       view_limits limits{{0.0, 0.0},
+                          {Axis == axis::x ? 0.0 : full_extent,
+                           Axis == axis::x ? full_extent : 0.0}};
+       for (std::size_t i = 0; i != tile.size();  ++i)
+         {
```
djowel, line 101: 3 space indents and braces aligned with `for`

djowel: Mostly formatting stuff, and this is good to go.

djowel:
One thing though: we will have to the same with the `skia_2024` branch. Then, we'll have to merge `master` with `skia_2024` (`master` tracks `skia_2024`). I keep both merged and in sync with the objective that they will one day be united. Obviously, there will be conflicts all over the place. I can do the `master` -> `skia_2024` merge if you can do the port to `skia_2024`.

bruno-j-nicoletti:
I can't get onto this again for a few days. The `skia_2024` modifications would mean applying the rect/point changes to the rect/point classes in the `artist` module wouldn't it and refactoring the tile classes appropriately?

lib/include/elements/support/rect.hpp
bruno-j-nicoletti: Fine.

djowel:
> I can't get onto this again for a few days. The `skia_2024` modifications would mean applying the rect/point changes to the rect/point classes in the `artist` module wouldn't it and refactoring the tile classes appropriately?

I suspect all that is needed is to apply the rect/point changes. The tiles will use that automatically by a simple `using artist::rect` and `using artist::point` which should already be there.

djowel:
In the meantime, I changed the target to a feature branch `tile-refactoring`, so I can continue with documentation which requires regular sync from `skia_2024` and `master`.

djowel:
OK to merge so far?

bruno-j-nicoletti:
> OK to merge so far?

I haven't fixed the formatting yet.

bruno-j-nicoletti:
Just made the last set of changed you suggested. Feel free to merge.

PR #417, opened 2024-06-28
busy_bar improvements

Flole998:
busy_bar: Make width of animation configurable
busy_bar: Fix issue when calling start twice

### DIFF
--- a/lib/include/elements/element/status_bar.hpp
+++ b/lib/include/elements/element/status_bar.hpp
@@ -100,11 +100,15 @@ namespace cycfi::elements
                               )
                                : status_bar_base(init_value)
                                , _start_pos(start_pos)
-                               , _status(-0.2)
+                               , _animation_width(-0.2)
+                               , _status(_animation_width)
                               {}
 
       void                    start_pos(double val);
       double                  start_pos() const { return _start_pos; }
+
+      void                    animation_width(double val);
+      double                  animation_width() const { return _animation_width; }
 
       void                    start(view& view_, duration time);
       void                    stop(view& view_);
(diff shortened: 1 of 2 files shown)


lib/include/elements/element/status_bar.hpp
```diff
@@ -117,6 +121,7 @@ namespace cycfi::elements
       void                    animate(view& view_);
 
       double                  _start_pos;
+      double                  _animation_width;
```
djowel, line 124: I've been thinking about `double` vs `float` since my last refactor. Perhaps we need to refactor these indicators to use `float` instead? `float` ought to be enough to save space, esp. now that we have more.

lib/src/element/status_bar.cpp
```diff
@@ -92,6 +92,11 @@ namespace cycfi::elements
       _start_pos = clamp(val, 0.0, 1.0);
    }
 
+   void busy_bar_base::animation_width(double val)
```
djowel, line 95: I've also been mulling about using the "styler" concept (see the buttons, for example) as a more adaptive design strategy, but that can come later.

lib/include/elements/element/status_bar.hpp
Flole998: It will only be a value between 0 and 1, so a float should be sufficient.

djowel:
Merged, but the example is not working on first click (does not do anything on first click).

djowel:
> Merged, but the example is not working on first click (does not do anything on first click).

Fixed. 

I think animation width should be a theme variable, though. You'd generally want consistency per app.

PR #431, opened 2024-10-31
adjust the position of the input method window so that it follows the caret

zhkkjun:
before
图片

after
图片

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -348,9 +348,10 @@ if (WIN32 AND NOT MSVC)
    target_link_libraries(elements PUBLIC ws2_32)
 endif()
 
-# Replace missing Shcore library removed by WIN32_LEAN_AND_MEAN
+# 1.Replace missing Shcore library removed by WIN32_LEAN_AND_MEAN
+# 2.IME requires Imm32
 if (WIN32)
-   target_link_libraries(elements PUBLIC Shcore)
+   target_link_libraries(elements PUBLIC Shcore Imm32)
 endif()
 
 add_library(cycfi::elements ALIAS elements)
--- a/lib/host/windows/key.cpp
+++ b/lib/host/windows/key.cpp
@@ -72,7 +72,7 @@ namespace cycfi::elements
       {
          // IME notifies that keys have been filtered by setting the virtual
          // key-code to VK_PROCESSKEY
-         return key_code::unknown;
+         return key_code::ime_process_key;
       }
 
       unsigned int key = HIWORD(lparam) & 0x1FF;
@@ -206,5 +206,6 @@ namespace cycfi::elements
          case 0x037:  return key_code::kp_multiply;
          case 0x04a:  return key_code::kp_subtract;
       }
+      return key_code::unknown;
    }
 }
--- a/lib/include/elements/base_view.hpp
+++ b/lib/include/elements/base_view.hpp
@@ -144,6 +144,9 @@ namespace cycfi::elements
 
    enum class key_code : int16_t
    {
+#ifdef WIN32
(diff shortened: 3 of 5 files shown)


djowel:
Can you explain a bit more? I have some concerns about the code changes. Thank you!

zhkkjun:
大体流程是监听窗体的键盘消息识别出输入法的虚拟键码VK_PROCESSKEY，然后在basic_text_box::key中识别到这个键码，计算出插入位置的坐标，然后获取窗体的输入法上下文，设置输入法的候选框和合成框的位置。
    根据之前的代码，VK_PROCESSKEY 会被识别成 key_code::unknown，后续不会进入“basic_text_box::key”中，变更后会进入“basic_text_box::key”中，确实会对之前的代码流程产生影响，不过 VK_PROCESSKEY  本身不会导致输入位置发生变化，因此不会产生实际影响。另外枚举key_code::ime_process_key设置为-2可能也不是很合适。
    总之这里仅是提供一个思路，你可以对代码做出合适的调整

Below is the AI translation:
    The process involves monitoring the keyboard messages of the window to identify the virtual key code VK_PROCESSKEY of the input method. Then, in “basic_text_box::key” , recognize this key code, calculate the coordinates of the insertion point, obtain the input context of the window, and set the position of the input method's candidate window and composition window.
    Based on the previous code, VK_PROCESSKEY would be recognized as key_code::unknown, and subsequently, it would not enter basic_text_box::key. After the change, it will enter basic_text_box::key. This indeed affects the previous code flow. However, VK_PROCESSKEY itself does not cause the input position to change, so there should be no practical impact. Additionally, setting key_code::ime_process_key to -2 may not be entirely appropriate. 
    In summary, this is just a suggestion, and you can make suitable adjustments to the code.

djowel:
Thank you for your contribution. I appreciate it!

Is it possible to move all platform-specific (Windows) code to `lib/host/windows/` ?  In particular, `lib/include/elements/base_view.hpp`, `lib/include/elements/element/text.hpp`, `lib/src/element/text.cpp` should not contain any platform specific code.  All platform-specific code should be in the `host`.  Such functions could be abstracted out with platform independent code.

Also, how about Linux and MacOS?

zhkkjun:
> Thank you for your contribution. I appreciate it!
> 
> Is it possible to move all platform-specific (Windows) code to `lib/host/windows/` ? In particular, `lib/include/elements/base_view.hpp`, `lib/include/elements/element/text.hpp`, `lib/src/element/text.cpp` should not contain any platform specific code. All platform-specific code should be in the `host`. Such functions could be abstracted out with platform independent code.
> 
> Also, how about Linux and MacOS?

yes, I will make some adjustments to the code later. And I tested the Linux platform and found that the input method cannot be invoked, this issue likely applies to Japanese and Korean input as well.There are several input method frameworks on Linux, so corresponding adaptations will likely be needed.

djowel:
> > Thank you for your contribution. I appreciate it!
> > Is it possible to move all platform-specific (Windows) code to `lib/host/windows/` ? In particular, `lib/include/elements/base_view.hpp`, `lib/include/elements/element/text.hpp`, `lib/src/element/text.cpp` should not contain any platform specific code. All platform-specific code should be in the `host`. Such functions could be abstracted out with platform independent code.
> > Also, how about Linux and MacOS?
> 
> yes, I will make some adjustments to the code later. And I tested the Linux platform and found that the input method cannot be invoked, this issue likely applies to Japanese and Korean input as well.There are several input method frameworks on Linux, so corresponding adaptations will likely be needed.

Wonderful! Let's keep this open. I want to merge this when it becomes mature and can support all supported OSes.

Issue #443, opened 2025-11-10
Selection list element doesn't call "on_select" with keys


tmccartney:
Hi, Thankyou for the great library. 

This may be a design decision or misunderstanding on my part; however, I thought I would raise an issue in case it is an oversight.

The selection_list_element handles key_code::up and key_code::down to change the selection but never calls the on_select function.

It seems to me that on_select should be called when selection changes from a key event (in selection_list_element::key), similar to the behaviour of a click event (in selection_list_element::click).

PR #444, opened 2025-11-11
Selection keys fix - calling on_select for keys selection events

tmccartney:
Proposed solution to #443

I noticed the Ctrl+a (select all) shortcut was also not calling on_select so have proposed a solution there too - which copies the selection_list_element::select_all() method. With a check included for if the selection_list_element has multi_select enabled. I am assuming the desired Ctrl+a behaviour with multi-select disabled is do nothing.


djowel:
Thank you!

PR #446, opened 2025-11-23
Update Asio to 1.36.0 and refactor io_context usage

### LINKED ISSUE #445: Support asio >= 1.33
Currently elements can't build with asio >= 1.33 cause asio removed some previously deprecated facilities in [1.33](https://www.boost.org/doc/libs/latest/doc/html/boost_asio/history.html)

For example, **Removed deprecated class io_context::work**, so the following code will cause a build error when using asio >= 1.33
https://github.com/cycfi/elements/blob/28ede99e37597d743979b127191df45ace11f58b/lib/include/elements/view.hpp#L148

miyanyan:
Fixes https://github.com/cycfi/elements/issues/445
Bump Asio dependency from 1.29.0 to 1.36.0 in CMake configuration. 
Refactor usage of io_context::work to asio::executor_work_guard and replace io().post with asio::post for improved compatibility with newer Asio versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ xcuserdata/
 /CMakeSettings.json
 /docs/antora/node_modules/
 /docs/antora/build/
+/.cache/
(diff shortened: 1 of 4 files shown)


djowel:
Merged. Thanks!

